PR #3, opened 2024-01-03
Added connected components

RyanXXYC:


src/connectComp.cpp
```diff
@@ -0,0 +1,265 @@
...
+	int y;
+};
+
+struct Edge {
+	int magnitude; // number of points in this edge
+	std::vector<Point> points;
+	// int Xmin, Xmax, Ymin, Ymax;
+};
+
+std::vector<Edge> NaiveConnectedComponent(unsigned char *image, int imageWidth, int imageHeight) { // const??
```
nguy8tri, line 28: You may go ahead and const the image field, but you should do this at both the pointer level and the value level. This will make sure that you cannot change neither the value nor the reference stored by this pointer. If its inconvenient, it does not need to be done though.

src/connectComp.cpp
```diff
@@ -0,0 +1,265 @@
...
+	int threshold = 3; // over threshold means possible components exist
```
nguy8tri, line 29: This is fine for now, I will put in an edit to this particular line when Karuna pushes her algorithm. In the meantime, please place a `TODO: Edit this for compatibility with edge filtering` comment above this

src/connectComp.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+	// Tells us if certian numbers in the below array are part of the same component
+    std::unordered_map<int, int> equivalencies;
+
+	// make an array with the same dimensions of the image, but it shows connected components using numbers
+    unsigned char *edges = (unsigned char *) std::malloc(imageWidth * imageHeight * sizeof(unsigned char));
```
nguy8tri, line 35: Here, we might be better off with a stack allocated array instead, since it isn't really relevant across the entire program.

src/connectComp.cpp
```diff
@@ -0,0 +1,265 @@
...
+    edges[0] = (image[0] > threshold) ? 1 : 0;
+    std::cout << static_cast<int>(edges[0]) << std::endl;
```
nguy8tri, line 37: I bet this print statement shouldn't be here. Eventually, when you're done debugging and testing, I will ask that you remove all print statements that shouldn't exist in the program, as well as "commented-out" code.

src/connectComp.cpp
```diff
@@ -0,0 +1,265 @@
...
+                }
+            }
+        }
+        else {
+            edges[i] = 0;
+        }
+    }
+    std::cout << std::endl;
+
+    for (int i = 0; i < imageHeight; ++i) {
```
nguy8tri, line 147: Another debugging tool here

src/connectComp.cpp
```diff
@@ -0,0 +1,265 @@
...
+    //         std::cout << "(" << static_cast<int>(edges[i * imageWidth + j]) << ") ";
+    //     }
+    //     std::cout << std::endl;
+    // }
+    
+
+	// Initial labeling of centroids
+    for(long i = 1; i < imageHeight * imageWidth; i++) {
+        // This is specific to the centroiding algo, which tests if this is even part of a centroid
+        if(image[i] > threshold) {
```
nguy8tri, line 51: Above, comment with: `TODO: Edit this for compatibility with edge filtering`

src/connectComp.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+std::vector<Edge> NaiveConnectedComponent(unsigned char *image, int imageWidth, int imageHeight) { // const??
+	int threshold = 3; // over threshold means possible components exist
+
+	// Tells us if certian numbers in the below array are part of the same component
+    std::unordered_map<int, int> equivalencies;
+
+	// make an array with the same dimensions of the image, but it shows connected components using numbers
+    unsigned char *edges = (unsigned char *) std::malloc(imageWidth * imageHeight * sizeof(unsigned char));
+    edges[0] = (image[0] > threshold) ? 1 : 0;
```
nguy8tri, line 36: Above, comment with: `TODO: Edit this for compatibility with edge filtering`

src/connectComp.cpp
```diff
@@ -0,0 +1,265 @@
...
+#include <vector>
+#include <iostream>
+
+#include <unordered_map>
+
+#include <unordered_set>
+#include <string.h>
+#include <cassert>
+#include <cstdlib>
+
```
nguy8tri, line 14: `#include` your .hpp file here. You'll then notice that there is a double definition, but the below comments will fix that.

src/connectComp.hpp
```diff
@@ -0,0 +1,25 @@
...
+    int y;
+};
+
+struct Edge {
+    int magnitude; // number of points in this edge
+    std::vector<Point> points;
+};
+
+// Function prototype
+std::vector<Edge> NaiveConnectedComponent(unsigned char *image, int imageWidth, int imageHeight);
```
nguy8tri, line 21: Please put a more descriptive function comment here. There should be:
1) General Description of the Funciton
2) Parameter Description
3) Outputs
4) Any modifications on the parameters
5) Preconditions
6) (optional) Postconditions

src/connectComp.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+}
+
+#endif
```
nguy8tri, line 25: I know you have completed tests on this function, so I will need you to formalize that in the test suite. That can be found in "found/test/<nameOfThisFile>test.cpp" (No .hpp file needed here). `example.cpp` will help you figure out how to write tests using C++'s CHECK framework (I'm open to other test suites if its better by the way).

nguy8tri (commented):
Great work Ryan! The review here is going to just be a stylistic review, and once you're done making the changes, I will then review your code for correctness.

PR #4, opened 2024-05-02
Added distance.cpp for the SphericalDistanceDeterminationAlgorithm

tan7271:
-Changed a few header files (namely attitude-utils) to add midpoint functions for Vec2 and Vec3
-Line 49 of Distance.cpp python code was `center = np.linalg.solve(A, y)`, which i used the inverse matrix function to solve for instead.
(Probably contains a few errors, if you see them feel free to point out)

### DIFF
--- a/src/attitude-utils.cpp
+++ b/src/attitude-utils.cpp
@@ -498,7 +498,7 @@ decimal AngleUnit(const Vec3 &vec1, const Vec3 &vec2) {
 }
 
 /**
- * Determines the  between two vectors
+ * Determines the Distance between two vectors
  * 
  * @param v1 The first vector
  * @param v2 The second vector
--- a/src/attitude-utils.cpp
+++ b/src/attitude-utils.cpp
@@ -426,6 +426,47 @@ Vec3 Vec3::operator*(const Mat3 &other) const {
 ///////////////////////////////////
 
 /**
+* Finds the midpoint between two different vectors
+*
+* @param vec1 The first vector
+* @param vec2 The second vector
+*
+* @return The mid point vector
+*
+*/
+Vec2 midpoint(const Vec2 &vec1, const Vec2 &vec2){
+    return {(vec1.x + vec2.x)/2, (vec1.y + vec2.y)/2};
+}
+
+/**
+* Finds the midpoint between two different vectors
+*
+* @param vec1 The first vector
+* @param vec2 The second vector
+*
+* @return The mid point vector
+*
+*/
+Vec3 midpoint(const Vec3 &vec1, const Vec3 &vec2){
+    return {(vec1.x + vec2.x)/2, (vec1.y + vec2.y)/2, (vec1.z + vec2.z)/2};
+}
+
+/**
+* Finds the midpoint between three different vectors
+*
+* @param vec1 The first vector
+* @param vec2 The second vector
+* @param vec3 The third vector
+*
+* @return The mid point vector
+*
+*/
+Vec3 midpoint(const Vec3 &vec1, const Vec3 &vec2, const Vec3 &vec3){
+    return {(vec1.x + vec2.x + vec3.x)/3, (vec1.y + vec2.y + vec3.y)/3, (vec1.z + vec2.z + vec3.z)/3};
+}
+
+
+/**
  * Determines the angle between two different vectors
  * 
  * @param vec1 The first vector
@@ -457,7 +498,7 @@ decimal AngleUnit(const Vec3 &vec1, const Vec3 &vec2) {
 }
 
 /**
- * Determines the distance between two vectors
+ * Determines the  between two vectors
  * 
  * @param v1 The first vector
  * @param v2 The second vector
@@ -467,6 +508,19 @@ decimal AngleUnit(const Vec3 &vec1, const Vec3 &vec2) {
 */
 decimal Distance(const Vec2 &v1, const Vec2 &v2) {
     return sqrt(pow(v1.x-v2.x, 2) + pow(v1.y-v2.y, 2));
+}
+
+/**
+ * Determines the  between two vectors
+ * 
+ * @param v1 The first vector
+ * @param v2 The second vector
+ * 
+ * @return The distance between v1 and v2
+ * 
+*/
+decimal Distance(const Vec3 &v1, const Vec3 &v2) {
+    return sqrt(pow(v1.x-v2.x, 2) + pow(v1.y-v2.y, 2) + pow(v1.z-v2.z, 2));
 }
 
 ///////////////////////////////////
--- a/src/attitude-utils.hpp
+++ b/src/attitude-utils.hpp
@@ -104,6 +104,7 @@ class Mat3 {
 public:
     decimal x[9];
 
+    
     // Accessor
 
     decimal At(int i, int j) const;
@@ -223,6 +224,11 @@ private:
     Mat3 dcm; // direction cosine matrix
     AttitudeType type;
 };
+
+// Vector operations
+Vec2 midpoint(const Vec2 &, const Vec2 &);
+Vec3 midpoint(const Vec3 &, const Vec3 &);
+Vec3 midpoint(const Vec3 &, const Vec3 &, const Vec3);
 
 // DCM-Quaternion-Spherical Conversions
 
--- a/src/distance.cpp
+++ b/src/distance.cpp
@@ -1,0 +1,81 @@
+#include "attitude-utils.hpp"
+#include "style.hpp"
+#include "camera.hpp"
+#include "distance.hpp"
+#include <math.h>
+
+namespace found {
+    class SphericalDistanceDeterminationAlgorithm {
+    public:         
+        distFromEarth Run(char *image, Points &p, int imageWidth, int imageHeight)
+        {
+            return solve(p, radius_);
+        }
+    private:
+
+        Vec3 getCenter(Vec3* spats){
+            Vec3 diff1 = spats[1] - spats[0];
+            Vec3 diff2 = spats[2] -spats[1];
+
+            Vec3 circleN = diff1.CrossProduct(diff2);
+            Vec3 circlePt = spats[0];
+
+            Vec3 mid1 = midpoint(spats[0], spats[1]);
+            Vec3 mid2 = midpoint(spats[1], spats[2]);
+
+            Vec3 mid1N = diff1;
+            Vec3 mid2N = diff2;
+
+            Mat3 matrix;
+            matrix.x[0] = circleN.x;
+            matrix.x[1] = circleN.y;
+            matrix.x[2] = circleN.z;
+            matrix.x[3] = mid1N.x;
+            matrix.x[4] = mid1N.y;
+            matrix.x[5] = mid1N.z;
+            matrix.x[6] = mid2N.x;
+            matrix.x[7] = mid2N.y;
+            matrix.x[8] = mid2N.z; 
+
+            int alpha = circleN*circlePt;
+            int beta = mid1N*mid1;
+            int gamma = mid2N*mid2;
+
+            Vec3 y;
+            y.x = alpha;
+            y.y = beta;
+            y.z = gamma;
+
+            Vec3 center = y * matrix.Inverse();
+
+            return center;
+        }
+
+        decimal getRadius(Vec3* spats, Vec3 center){
+            return Distance(spats[0], center);
+        }
+
+        decimal getDistance(decimal r)
+        {
+            return 6378/r;
+        }
+
+        distFromEarth solve(Points& pts, int R){
+            Vec3 spats[3] = {cam_.CameraToSpatial(pts[0]), cam_.CameraToSpatial(pts[1]), cam_.CameraToSpatial(pts[2])};
+            Vec3 uSpats[3] = {spats[0].Normalize(), spats[1].Normalize(), spats[2].Normalize()};
+
+            Vec3 center = getCenter(uSpats);
+            decimal r = getRadius(uSpats, center);
+
+            decimal h = getDistance(r);
+
+            printf("========= solution =============\n");
+            printf("((center), h) = ((%f, %f, %f), %f)\n", center.x,center.y,center.z,r);
+            printf("Distance from surface: %f\n", h - R);
+
+            return center * h;
+        }
+
+        
+    };
+}
--- a/src/distance.hpp
+++ b/src/distance.hpp
@@ -2,6 +2,8 @@
 #define DISTANCE_H
 
 #include "style.hpp"
+#include "attitude-utils.hpp"
+#include "camera.hpp"
 
 namespace found {
 
@@ -25,7 +27,7 @@ public:
      * 
      * @return The distance of the satellite from Earth
     */
-    virtual distFromEarth Run(char* image, Points &p /*More go here*/) = 0;
+    virtual distFromEarth Run(char* image, Points &p, int imageWidth, int imageHeight /*More go here*/) = 0;
 };
 
 /**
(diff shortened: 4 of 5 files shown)


src/attitude-utils.cpp
```diff
@@ -425,6 +425,47 @@ Vec3 Vec3::operator*(const Mat3 &other) const {
...
 ///////////////////////////////////
 
+/**
+* Finds the midpoint between two different vectors
+*
+* @param vec1 The first vector
+* @param vec2 The second vector
+*
+* @return The mid point vector
+*
```
nguy8tri, line 435: As a convention, the there should be no space between the last content and the `*/` character (remove this space everywhere)

src/attitude-utils.cpp
```diff
@@ -469,6 +510,19 @@ decimal Distance(const Vec2 &v1, const Vec2 &v2) {
     return sqrt(pow(v1.x-v2.x, 2) + pow(v1.y-v2.y, 2));
 }
 
+/**
+ * Determines the  between two vectors
```
nguy8tri, line 514: Extra Space here

src/attitude-utils.cpp
```diff
@@ -469,6 +510,19 @@ decimal Distance(const Vec2 &v1, const Vec2 &v2) {
...
+ * 
+ * @param v1 The first vector
+ * @param v2 The second vector
+ * 
+ * @return The distance between v1 and v2
+ * 
+*/
+decimal Distance(const Vec3 &v1, const Vec3 &v2) {
+    return sqrt(pow(v1.x-v2.x, 2) + pow(v1.y-v2.y, 2) + pow(v1.z-v2.z, 2));
```
nguy8tri, line 523: Another way we can do this, is to compute the magnitude of the difference between the two vectors.

So, think about which one is faster & more memory efficient.

src/attitude-utils.hpp
```diff
@@ -104,6 +104,7 @@ class Mat3 {
 public:
     decimal x[9];
 
+
```
nguy8tri, line 107: Let's not add this space

src/attitude-utils.hpp
```diff
@@ -104,6 +104,7 @@ class Mat3 {
 public:
     decimal x[9];
 
+    
     // Accessor
```
nguy8tri, line 109: Delete Space

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
+#include "attitude-utils.hpp"
+#include "style.hpp"
+#include "camera.hpp"
+#include "distance.hpp"
+#include <math.h>
+
```
nguy8tri, line 6: Your includes should be in a specific order, and seperated by the hierarchy of categories:
- Any C/C++ standard libraries (with are surrounded by <>)
- Any External libraries that you're using (doesn't apply here, but would be something like Python's numpy)
- Any libraries that you've defined outside of this folder (also doesn't apply here)
- Any libraries that you've defined inside this folder that are not this header file (e.g. style.hpp)
- The header file that corresponds to this one
```
#include <math.h>

#include "attitude-utils.hpp"
#include "style.hpp"
#include "camera.hpp"

#include "distance.hpp"
```

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+namespace found {
+    class SphericalDistanceDeterminationAlgorithm {
```
nguy8tri, line 8: You should change the below headers to not be enclosed in the class (still in namespace). In order to change all of this then, you're going to have change the definition format. (You don't need to specify what's public and what's private).

Every method should now be declared like the following:
`returnType ClassName::MethodName(params...)`

As an example:
`distFromEarth SphericalDistanceDeterminationAlgorithm::Run(char *image, Points &p, int imageWidth, int imageHeight)`

Also, do not indent within a namespace

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+            printf("========= solution =============\n");
+            printf("((center), h) = ((%f, %f, %f), %f)\n", center.x,center.y,center.z,r);
+            printf("Distance from surface: %f\n", h - R);
+
+            return center * h;
+        }
+
+        
+    };
+}
```
nguy8tri, line 81: You need a semicolon after the `}` which ends the namespace

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+            matrix.x[0] = circleN.x;
+            matrix.x[1] = circleN.y;
+            matrix.x[2] = circleN.z;
+            matrix.x[3] = mid1N.x;
+            matrix.x[4] = mid1N.y;
+            matrix.x[5] = mid1N.z;
+            matrix.x[6] = mid2N.x;
+            matrix.x[7] = mid2N.y;
+            matrix.x[8] = mid2N.z; 
+
```
nguy8tri, line 39: For readibility, you might be able to do something like `matrix.x = circleN`, and then `(matrix.x + 3) = mid1N` and so on

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+            int alpha = circleN*circlePt;
+            int beta = mid1N*mid1;
+            int gamma = mid2N*mid2;
+
+            Vec3 y;
```
nguy8tri, line 44: A way to intitialize this y up here, is to do:
`Vec3 y = {alpha, beta, gamma}`

src/distance.cpp
nguy8tri: You might be able to actually just do `matrix = {...}`

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+            y.x = alpha;
+            y.y = beta;
+            y.z = gamma;
+
+            Vec3 center = y * matrix.Inverse();
+
```
nguy8tri, line 50: I bet what you're going for here is the answer to:
$$Ax = b$$

But, what you want here is:
$$A^{-1}Ax=A^{-1}b$$
$$x=A^{-1}b$$

The above also doesn't work because the dimensions don't line up. Taking the inner product of any two matricies has to be, such that we have two matricies A and B, A has to have the same number of columns as the rows in B, such that if the shapes are A is (n, k), and B is (k, m):
$$(n, k) * (k, m) \rightarrow (n, m)$$

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+            return center;
+        }
+
+        decimal getRadius(Vec3* spats, Vec3 center){
+            return Distance(spats[0], center);
+        }
+
+        decimal getDistance(decimal r)
+        {
+            return 6378/r;
```
nguy8tri, line 60: Use a macro to define this "magic number" so that know what it means.

The define macro basically defines a "constant" in your system.
`#define RADIUS_OF_EARTH 6378.0`

Technically, Google Style Guides don't like this, so if you want to do something else, then use the following:
`static constexpr decimal radius_of_earth = 6378.0`

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+        distFromEarth solve(Points& pts, int R){
+            Vec3 spats[3] = {cam_.CameraToSpatial(pts[0]), cam_.CameraToSpatial(pts[1]), cam_.CameraToSpatial(pts[2])};
+            Vec3 uSpats[3] = {spats[0].Normalize(), spats[1].Normalize(), spats[2].Normalize()};
+
+            Vec3 center = getCenter(uSpats);
+            decimal r = getRadius(uSpats, center);
+
+            decimal h = getDistance(r);
+
+            printf("========= solution =============\n");
```
nguy8tri, line 72: When you're done, delete these print statements

src/distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        decimal getDistance(decimal r)
+        {
+            return 6378/r;
+        }
+
+        distFromEarth solve(Points& pts, int R){
+            Vec3 spats[3] = {cam_.CameraToSpatial(pts[0]), cam_.CameraToSpatial(pts[1]), cam_.CameraToSpatial(pts[2])};
+            Vec3 uSpats[3] = {spats[0].Normalize(), spats[1].Normalize(), spats[2].Normalize()};
```
nguy8tri, line 65: Go straight to uSpats and avoid defining spats

src/distance.hpp
```diff
@@ -36,15 +38,30 @@ class DistanceDeterminationAlgorithm {
 */
 class SphericalDistanceDeterminationAlgorithm : public DistanceDeterminationAlgorithm {
 public:
-    SphericalDistanceDeterminationAlgorithm(float radius);
+    SphericalDistanceDeterminationAlgorithm(float radius, Camera &cam) : radius_(radius), cam_(cam) {};
+    /**
```
nguy8tri, line 42: Delete this comment.

its equivalent to just using the constructor inside the `{}`

src/distance.hpp
```diff
@@ -36,15 +38,30 @@ class DistanceDeterminationAlgorithm {
...
-    distFromEarth Run(char* image, Points &p/*More go here*/) override;
+    distFromEarth Run(char* image, Points &p, int imageWidth, int imageHeight/*More go here*/) override;
 private:
     // Fields specific to this algorithm, and helper methods
+    Vec3 getCenter(Vec3* spats);
+    decimal getRadius(Vec3* spats, Vec3 center);
+    decimal getDistance(decimal r);
+    void solve(Points& pts, int R);
+    
+    Camera cam_;
```
nguy8tri, line 62: I wanna know what these fields are for, so place comments over them

src/style.hpp
```diff
@@ -16,7 +16,7 @@ typedef std::vector<Vec2> Points;
 
 // The output for Distance Determination Algorithms (distance.hpp/cpp). Currently 
 // set to a floating point value that represents the distance from Earth
```
nguy8tri, line 18: Change this description for us, so its now just a Vec3 thing instead of a floating point value

src/attitude-utils.cpp
```diff
@@ -425,6 +425,47 @@ Vec3 Vec3::operator*(const Mat3 &other) const {
 ///// VECTOR UTILITY FUNCTIONS ////
 ///////////////////////////////////
 
+/**
+* Finds the midpoint between two different vectors
+*
+* @param vec1 The first vector
+* @param vec2 The second vector
+*
+* @return The mid point vector
```
nguy8tri, line 434: It's midpoint not mid point

nguy8tri (commented):
Great job, just some changes before we test!

src/distance.cpp
```diff
@@ -0,0 +1,64 @@
+#include <math.h>
```
nguy8tri, line 1: Make any lines longer than 80 characters multiple lines! Check out cpplint.py to help you do this (extension in VS Code) if you want an easy way to do this

src/distance.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace found {  
+
+distFromEarth SphericalDistanceDeterminationAlgorithm::Run(char *image, Points &p, int imageWidth, int imageHeight)
+{
+    return solve(p, radius_);
+}
+
+Vec3 SphericalDistanceDeterminationAlgorithm::getCenter(Vec3* spats){
+    Vec3 diff1 = spats[1] - spats[0];
+    Vec3 diff2 = spats[2] -spats[1];
```
nguy8tri, line 20: space

src/distance.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    Vec3 circleN = diff1.CrossProduct(diff2);
```
nguy8tri, line 22: use std::move here

nguy8tri: Basically, whenever you assign an expression that is an object, use `std::move`

src/distance.hpp
nguy8tri: Please do this commenting for your private fields!

nguy8tri (commented):
Great job, just a few more styling issues!

src/attitude-utils.cpp
```diff
@@ -457,16 +490,27 @@ decimal AngleUnit(const Vec3 &vec1, const Vec3 &vec2) {
...
  * 
  * @param v1 The first vector
  * @param v2 The second vector
  * 
  * @return The distance between v1 and v2
- * 
 */
 decimal Distance(const Vec2 &v1, const Vec2 &v2) {
-    return sqrt(pow(v1.x-v2.x, 2) + pow(v1.y-v2.y, 2));
+    return (v1-v2).Magnitude();
```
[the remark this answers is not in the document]
tan7271: Didn't change, I think a rvalue is returned here so it is okay.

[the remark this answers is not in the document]
tan7271: When it comes to breaking it up these lines, is there a standard way to do it? I searched up the google cpp style guide and there wasn't an answer.

`decimal SphericalDistanceDeterminationAlgorithm::getRadius(Vec3* spats,`
`Vec3 center) {
    return Distance(spats[0], center);
}`

Is something like this okay?

src/distance.cpp
tan7271: Is this what you meant?

`Vec3 SphericalDistanceDeterminationAlgorithm::getCenter(Vec3* spats) {
    Vec3 diff1 = std::move(spats[1] - spats[0]);
    Vec3 diff2 = std::move(spats[2] - spats[1]);

    Vec3 circleN = std::move(diff1.CrossProduct(diff2));
    Vec3 circlePt = spats[0];

    Vec3 mid1 = std::move(midpoint(spats[0], spats[1]));
    Vec3 mid2 = std::move(midpoint(spats[1], spats[2]));

    Vec3 mid1N = std::move(diff1);
    Vec3 mid2N = std::move(diff2);

    Mat3 matrix;
    matrix = {circleN.x, circleN.y, circleN.z, mid1N.x, mid1N.y,
              mid1N.z, mid2N.x, mid2N.y, mid2N.z};

    int alpha = circleN*circlePt;
    int beta = mid1N*mid1;
    int gamma = mid2N*mid2;

    Vec3 y = {alpha, beta, gamma};

    Vec3 center = std::move(matrix.Inverse() * y);

    return center;
}`

nguy8tri:
Just merged from main into this branch (took about an hour, you're welcome). I also did some light debugging of your code, and it turns out the math behind some of your original code was wrong, so I fixed that, and your base case test does work now (hooray!). As per the new changes (which you can [view here](https://www.notion.so/huskysat/Infrastructure-CI-CD-Tools-1084a4f5c8e080338a8ec6be317cf513?pvs=4), with [presentation here](https://washington.zoom.us/rec/share/sXaH6k4Y2HJxWDX5BpaaaOeTeg0HMtD6esxBsh-GtlvTUyrkXC34Wgfwqo-JNHoa.JbL5NQD60UcvY1iP)), these changes automate many parts of code checking, which lets me focus more on your code. This includes the checks that are performed on your PR.

tools/generator/src/__main__.py
```diff
@@ -169,12 +169,11 @@ def generate_points(
     # Step 4: Transform points into camera coordinate system
     camera_points = camera.to_camera_space(points)
     camera_center = camera.to_camera_space({center})
-
+    print(f"Position rel Camera: {camera_center.normalize() * position.norm}")
     # Step 5: Project Points into Camera
     image_center, *image_points = camera.spatial_to_camera(
         camera_center, *camera_points
     )
-
```
nguy8tri, line 177: Reinsert extra line

tools/generator/src/spatial/coordinate.py
nguy8tri, line 1: Please revert all changes in this file

src/distance/distance.cpp
```diff
@@ -0,0 +1,82 @@
+#include <cmath>
+#include <utility>
+#include <iomanip>
+
+#include "spatial/attitude-utils.hpp"
```
nguy8tri, line 5: Use `#include "common/spatial/attitude-utils.hpp"` and `#include "common/spatial/camera.hpp"`

nguy8tri (approved):
Great work on your code, I approve these changes! I'm going to let you do the honors and merge it yourself!

PR #5, opened 2024-09-13
Refractor Code Base for Implementation

nguy8tri:
This PR attempts to refractor the code base to a more sophisticated look:
1. The `src` and `test` directories can now support multi-directory compilation
2. All compilation artifacts are directed into a `build` folder
3. `Makefile` now supports a whole workflow chain that can be used to validate user changes
 a. Compilation (and setup)
 b. Testing (and setup), which automatically triggers tests AND within valgrind to check for memory leaks
 c. Google Style Checking for both `src` and `test` folders (ignored flags and custom settings contained within `CPPLINT.cfg`
 d. All main targets can be triggered independently
5. A github workflow is now added to do what `Makefile` does
6. Migrated testing framework from `check` to `gtest` and `gmock` (to support mocking was the main motivation here)

This edit passes manual workflow as specified within the Makefile


src/distance/vectorize.hpp
```diff
@@ -34,15 +33,15 @@ class VectorGenerationAlgorithm {
...
-
+ public:
     /**
      * Creates a LOSTVectorGenerationAlgorithm object
      * 
      * @param orientation The orientation of the satellite as determined by LOST
     */
-    LOSTVectorGenerationAlgorithm(Vec3 orientation/*Params to initialze fields for this object*/) : orientation(orientation) {};
-    
+    explicit LOSTVectorGenerationAlgorithm(Vec3 orientation/*Params to initialze fields for this object*/)
```
zeddie888, line 42: nit: comment inside paren looks a bit weird to me, though I've never seen this before, what's the reasoning?

nguy8tri: Oh, just to warn people about what they need to feed into here, but its going to change soon, so I won't edit it now

README.md
```diff
@@ -8,9 +8,16 @@ If you are using Windows, we recommend downloading Windows Subsystem for Linux,
 
 ## Linux Prerequisites
 FOUND runs on a Linux System that has many requirements. The following packages are required, with download instructions given in Ubuntu:
+- Git (`sudo apt install git`)
```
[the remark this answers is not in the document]
nguy8tri: I would like to but then also mac? So I feel it might be best to just leave this for people to do themselves. I was thinking of putting this in the makefile, but then I'd run into the same issue.

zeddie888 (approved):
lgtm, will look at gmock myself later.

PR #9, opened 2025-03-17
Refractor v3 (Command-Line, Logging, Makefile, Floating-Point Types, Others)

nguy8tri:
# Description
Made Refactoring Changes to Repository to Improve Code Readability and Streamline Tools. Here are the biggest changes:
1. Command-Line
  a) `main` and `found::main` is now more readable. Parsing is also split into specialized functions to avoid confusion and bring clarity to different stages.
  b) `options.hpp` Consolidates all definitions pertaining to permissible flags under particular algorithms.
  c) Starter tests are inserted to make coverage pass
2. Logging
  a) Added Basic Logging Capabilities to the program, which is controlled via macros for program efficiency
  b) The `make` macro `ENABLE_LOGGING` enables logging and the macro `LOGGING_LEVEL` allows you to set the level to `INFO`, `WARN` or `ERROR`
  c) Info, Warning, and Error Logs are defined as macros, currently set to standard output and error. They can be redefined to be different (such as a file), within each `#ifndef` statement. Putting it outside will have bad consequences.
2. `Makefile`
  a) Now caching will completely build `googletest` module in cache before transfer to build (so we don't have to build `googletest` every single `make clean && make`)
  b) Runs -O1 optimization on test suite to get rid of useless lines of code
  b) Doxygen now sends its stderr output into the cache, which is then quickly deleted (minor change)
3. Floating-Point Types (`decimal`)
  a) Refractored `decimal` into its own file (aka I stole from LOST)
4. Other Changes
  a) Reordered `struct Image` fields to optimize for internal fragmentation
  b) Updated `environment.yml` for Image Generator to accurately reflect Python requirements
  c) Updated `install.sh` and `Dockerfile` to reflect accurate dependency versions

# Artifacts for PR #9 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/9/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/9/merge/doxygen)


src/command-line/parser.cpp
```diff
@@ -0,0 +1,124 @@
...
+#define REQ_ASSIGN(options, prop, value, default) \
+    options.prop = (value);
+
+/**
+ * Assigns a flag with an optional value
+ * 
+ * @param options The options object that has the flag
+ * @param prop The name of the prop within options
+ * @param value The value to assign
+ * @param default THe default value
```
j4lando, line 43: @param default The default value

j4lando (approved):

PR #10, opened 2025-04-08
Data serialization tests

ozbot11:
# Description
Fixed some functionality and conventions
- data_structures.hpp => data_file.hpp
- Vec3 instead of Position
- Removed coord_system[32] definition
- Added EulerAngles field to relative_attitude struct
- Handled endianness with ntoh, ntohl, ntohs, hton, htonl, htons
- Added folder src/test/serialization

More changes too, as requested by @nguy8tri

# Artifacts for PR #10 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/10/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/10/merge/doxygen)


.github/workflows/build.yml
nguy8tri, line 1: This goes for all the files you've written, but as I messaged you separately, your files appear to be using CRLF instead of just LF. Please change this for all files, as we have a standard of using LF.

nguy8tri (commented):
Hello, for now, I'll need you to change all your line endings to LF (hold off merging with main). I must see all the changes you've made more clearly to provide an accurate evaluation, and I can only do that when the changes due to CRLF line endings are out of the way. Thanks!

.gitattributes
nguy8tri, line 1: We do not need this file, please remove it. Chances are that on your machine, the default line ending is CRLF. To change this in your github settings, please `git config --global core.autocrlf input`, which will make git use LF line endings.

src/serialization/data_file.cpp
nguy8tri, line 1: So, a big thing I'm noticing is that your PR is not passing all the checks. You will need to fix your code so that it does. An easy way to ensure this is to see if running `make` locally passes. However, its visible in your code that it won't for a few reasons:
0. Upon inspection of the checks, your test suite fails to compile and thus cannot be tested. Please ensure your code works as a courtesy, but nothing that cannot be simply fixed.
1. There is no documentation over function definitions (caught by the `doxygen_generate` target in `Makefile`). When you put this in, please put it into the `.hpp` file declaration of a function/class over putting in into the `.cpp` file.
2. There are some violations of the Google Styleguide (caught by the `google_stylecheck` and `google_stylecheck_test` targets in `Makefile`)

I apologize if I sound harsh here, I don't mean to. Even I'm not perfect at these, and it takes time to get used to. To fix these issues, run `make` and the errors will tell you what to fix.

src/serialization/data_file.cpp
```diff
@@ -0,0 +1,15 @@
+#include "serialization/data_file.hpp"
+#include <stdexcept>
+
+namespace found {
+
+    void DataFileHeader::validate() {
+        if (magic[0] != 'F' || magic[1] != 'O' || magic[2] != 'U' || magic[3] != 'N') {
```
nguy8tri, line 7: I think I know why you've chosen this magic string, but you might want to explain why this is (via documentation).

src/serialization/data_file.cpp
```diff
@@ -0,0 +1,15 @@
...
+            throw std::runtime_error("Magic value mismatch");
+        }
```
nguy8tri, line 9: I think its a good thing you're completing these checks, but the choice to throw an error might be a bad one, especially since this will go onto an embedded system. I'm not asking you to change it, I'd just like to see a rationale behind why we should throw an error and not do something else (for instance, just say we have a blank data file).

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,155 @@
+#include <fstream>
+#include <iostream>
+#include "serialization/serialization.hpp"
+#include "spatial/attitude-utils.hpp"
+
+using Vec3 = found::Vec3;
+
+namespace found {
+
+    // inline uint16_t htons(uint16_t v) {
```
nguy8tri, line 10: Please remove commented out code

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,155 @@
+#include <fstream>
+#include <iostream>
+#include "serialization/serialization.hpp"
+#include "spatial/attitude-utils.hpp"
+
+using Vec3 = found::Vec3;
```
nguy8tri, line 6: This isn't necessary since you've re-defined the `found` namespace

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,155 @@
...
+    // Deserialize DataFile from a file
+    DataFile deserialize(std::istream& stream) {
+        DataFile data;
+
+        data.relative_attitude = EulerAngles(0, 0, 0);
+        ntoh(data.relative_attitude);
+
+        data.header = readHeader(stream);
+
+        stream.read(reinterpret_cast<char*>(&data.relative_attitude), sizeof(data.relative_attitude));
```
nguy8tri, line 124: shouldn't it be this line, and then ntoh. I'm also not sure what line 119 is doing if we just override that information?

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+
+        for (auto& r : data.positions) {
```
nguy8tri, line 127: This line spooks me out a bit. Because we don't actually know how big data should be. My suggestion here is to actually reformat your datafile so instead, it is:
```
...
size_t record_len;
LocationRecord *record; // Heap Allocated (i.e. = new LocationRecord[record_len], deletion is delete[] record)
```
This will make your reading easier (and writing as well).

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,155 @@
...
+            LocationRecord record = r;
+            hton(record.position);
+            record.timestamp = htonl(record.timestamp);
+            stream.write(reinterpret_cast<const char*>(&record), sizeof(LocationRecord));
+        }
+    }
+
+    // Deserialize DataFile from a file
+    DataFile deserialize(std::istream& stream) {
+        DataFile data;
```
nguy8tri, line 117: If I'm not mistaken (with the edit to `records` as I indicate below`), I'm almost certain you can actually just read the whole thing into the struct, and then `ntoh` everything. This should make the reading more efficient.

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,155 @@
...
+        // Read the positions
+        data.positions.resize(data.header.num_positions);
+        stream.read(reinterpret_cast<char*>(data.positions.data()), data.positions.size() * sizeof(LocationRecord));
+
+        return data;
+    }
+
+    // Read only the header from a file
+    DataFileHeader readHeader(std::istream& stream) {
+        DataFileHeader header;
```
nguy8tri, line 141: See comment above about just loading directly into the struct and then converting to host order.

src/serialization/serialization.hpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+inline uint32_t ntohl(uint32_t v) {
+    return ((v & 0xFF000000) >> 24) |
+           ((v & 0x00FF0000) >> 8) |
+           ((v & 0x0000FF00) << 8) |
+           ((v & 0x000000FF) << 24);
+}
+
+// inline uint16_t htons(uint16_t v);
```
nguy8tri, line 43: Delete all commented code out

test/serialization/serialization-test.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+TEST_F(SerializationTest, IncorrectHeader) {
+    std::string incorrectCRCTestHeaderString;
+    std::istringstream incorrectCRCTestHeaderStream;
+    incorrectCRCTestHeaderString = std::string(reinterpret_cast<const char*>(incorrectCRCTestHeader), sizeof(incorrectCRCTestHeader));
+    incorrectCRCTestHeaderStream = std::istringstream(incorrectCRCTestHeaderString);
+    ASSERT_THROW(readHeader(incorrectCRCTestHeaderStream), std::runtime_error);
+}
+
+TEST_F(SerializationTest, RoundTripSerialization) {
```
nguy8tri, line 57: I'd like to see just another test or two like this. In the industry, you may often hear its nice to test the "happy paths" or the data paths that people will most often use, which is normal operation of this file.

test/serialization/serialization-test.cpp
```diff
@@ -0,0 +1,165 @@
...
+    out.write(reinterpret_cast<const char*>(&header), sizeof(header));
+
+    std::string buffer = out.str();
+    std::istringstream in(buffer);
+
+    // The CRC might pass, but the magic number should trigger a validation failure
+    EXPECT_THROW({
+        readHeader(in);
+    }, std::runtime_error);
+}
```
nguy8tri, line 164: I think I'd like to see a bit more test coverage here, please add tests for the following cases:
1. No positions (this is actually common)
2. No relative orientation (Think about what your code should do in this case, should we tolerate such a case?)
3. Nothing else except a header (Again, do you want to tolerate this case?)

src/serialization/data_file.hpp
```diff
@@ -0,0 +1,33 @@
...
+};
+
+struct LocationRecord {
+    Vec3 position;
+    uint64_t timestamp;
+};
+
+struct DataFile {
+    DataFileHeader header;
+    EulerAngles relative_attitude = EulerAngles(0, 0, 0);  // Attitude of the object
```
nguy8tri, line 27: I'm very sorry about this, but could you change this to a Quaternion Object (no need to initialize it).

src/serialization/data_file.hpp
```diff
@@ -0,0 +1,33 @@
...
+    std::vector<LocationRecord> positions;
```
nguy8tri, line 28: As I mention below, the use of a vector makes your serialization/deserialization process complicated. Usually in a data file, people will instead replace this with a field for the size of this array, and then literally an array of that size with all the entries. This makes serialization and deserialization a linear process (i.e. you just read/write the data from top to bottom without going back up) and there's no need for post processing on this end. Its actually no different from a std::vector, except you don't get access to its internal fields. You will also find that this makes the process easier in that all you have to do is take your struct, throw `htonl` at the whole thing, and then send the entire struct into a file (to deserialize, read the entire file into the struct, then throw `ntohl` at the whole thing).

nguy8tri: As a caveat, you usually want to serialize/deserialize the fields individually (one by one) into the file. This is because structs usually contain padding, the rules of which apparently change between some systems. If you wanted to be exact, you'd serialize/deserialize the fields one by one, with no padding of course. Be sure to use precise types (i.e. Types whose size does not change across systems. For instance `char` is always 1 byte, but `int` is sometimes 4 bytes or 8 bytes).

nguy8tri (changes requested):
Overall, great job handling everything. Once again, I must say that I'm quite impressed with the work you've done here, it certainly confused me the first time I did a project like this. However, I'm realizing its my fault for not telling you some of the conventions of deserialization/serialization from the start, so I hope my comments make up for that!

P.S. Our repository goes through strict checking via pull requests. You'll notice that your pull request at the moment has a red X next to it because it did not pass the checks. Generally, you'll want to fix this before requesting a review, but its not a big deal, just remember this for the future (I will be asking if all the checks pass next time before another review). Before you commit a change, you can see if it will pass simply by running `make` in your own local repository. If the command passes (in VS code, you can tell if a command passes if the light next to the line you ran the command is not red. Since we have a strict `Makefile`, you can tell if it passed more easily when you run `make` if you see the line `Doxygen completed successfully with no warnings.` as the last printed line).

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+/// Converts a DataFileHeader from host to network byte order.
+void hton(DataFileHeader& header) {
+    header.version = htonl(header.version);
+    header.num_positions = htonl(header.num_positions);
+    header.crc = htonl(header.crc);
+}
+
+/// Converts EulerAngles from host to network byte order.
+void hton(EulerAngles& angles) {
+    angles.roll = htond(angles.roll);
```
nguy8tri, line 19: `htondec`

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+    angles.ra = htond(angles.ra);
+    angles.de = htond(angles.de);
+}
+
+/// Converts a Vec3 from host to network byte order.
+void hton(Vec3& v) {
+    v.x = htond(v.x);
```
nguy8tri, line 26: `htondec`

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+/// Converts a DataFileHeader from network to host byte order.
+void ntoh(DataFileHeader& header) {
+    header.version = ntohl(header.version);
+    header.num_positions = ntohl(header.num_positions);
+    header.crc = ntohl(header.crc);
+}
+
+/// Converts EulerAngles from network to host byte order.
+void ntoh(EulerAngles& angles) {
```
nguy8tri, line 39: `ntohdec`

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+    angles.roll = ntohd(angles.roll);
+    angles.ra = ntohd(angles.ra);
+    angles.de = ntohd(angles.de);
+}
+
+/// Converts a Vec3 from network to host byte order.
+void ntoh(Vec3& v) {
```
nguy8tri, line 46: `ntohdec`

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+    return crc ^ 0xFFFFFFFFU;
+}
+
+/// Serializes a DataFile object to an output stream.
+void serialize(const DataFile& data, std::ostream& stream) {
+    DataFileHeader header = data.header;
+    hton(header);
+    header.crc = calculateCRC32(&header, sizeof(header) - sizeof(header.crc));
```
nguy8tri, line 72: swap line 72 and 71, then delete 73 (because it will already have been swapped)

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+            else
+                crc = crc >> 1;
+        }
+    }
+    return crc ^ 0xFFFFFFFFU;
+}
+
+/// Serializes a DataFile object to an output stream.
+void serialize(const DataFile& data, std::ostream& stream) {
+    DataFileHeader header = data.header;
```
nguy8tri, line 70: For the num_positions field, do we want to also calculate that in here, or make the user calculate it themselves. You should specify if the user needs to do it as a `@pre` tag in the comment above the function declaration in the `.hpp` file.

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+    header.crc = htonl(header.crc);
+    stream.write(reinterpret_cast<const char*>(&header), sizeof(header));
+
+    EulerAngles relative_attitude = data.relative_attitude;
+    hton(relative_attitude);
+    stream.write(reinterpret_cast<const char*>(&relative_attitude), sizeof(relative_attitude));
+
+    for (auto& r : data.positions) {
+        LocationRecord record = r;
+        hton(record.position);
```
nguy8tri, line 82: Its okay that you have this as is, but I'm just curious why we didn't encapsulate `hton` for a `LocationRecord`, but we did it for everything else.

src/serialization/data_file.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * @brief Relative orientation (attitude) of the object as Euler angles.
+     */
+    EulerAngles relative_attitude = EulerAngles(0, 0, 0);
+
+    /**
+     * @brief Collection of location records in the file.
+     */
+    std::vector<LocationRecord> positions;
```
nguy8tri, line 92: I want this to be a `std::unique_ptr<LocationRecord>`, not a `std::vector<LocationRecord>`

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+    DataFile data;
+    data.header = readHeader(stream);
+
+    stream.read(reinterpret_cast<char*>(&data.relative_attitude), sizeof(EulerAngles));
+    if (stream.gcount() != sizeof(EulerAngles)) {
+        throw std::ios_base::failure("Failed to read relative_attitude");
+    }
+    ntoh(data.relative_attitude);
+
+    data.positions.reserve(data.header.num_positions);
```
nguy8tri, line 99: After changing `data.positions` to `std::unique_ptr<LocationRecord>`, you can then change this line to be `data.positions = std::make_unique<LocationRecord[]>(data.header.num_positions)`;

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+    for (uint32_t i = 0; i < data.header.num_positions; ++i) {
+        LocationRecord record;
```
nguy8tri, line 101: Then, here, iterate through your array instead of creating a temporary copy, filling the data in each entry. So, roughly:

```
read(reinterpret_cast<char*>(&data.positions[i]), sizeof(LocationRecord));
// error check
ntoh(data.positions[i]);
```

which is simpler now

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,142 @@
...
+    hton(header);
+    header.crc = calculateCRC32(&header, sizeof(header) - sizeof(header.crc));
+    header.crc = htonl(header.crc);
+    stream.write(reinterpret_cast<const char*>(&header), sizeof(header));
+
+    EulerAngles relative_attitude = data.relative_attitude;
+    hton(relative_attitude);
+    stream.write(reinterpret_cast<const char*>(&relative_attitude), sizeof(relative_attitude));
+
+    for (auto& r : data.positions) {
```
nguy8tri, line 80: After changing to `LocationRecord *` you may not be able to for-each

nguy8tri (changes requested):
Great Job Erim, just a small number of changes I'm requesting this time around. But overall, you've grasped serialization/deserialization quite well, the code now looks like what I'm used to! Please also be sure to merge with main

As for the build process, you only have minor errors here, but I am letting you correct these. You will also have errors for documentation, as some members have not been documented:
```
Doxygen found errors and warnings while generating documentation:
/home/tri-nguyen/found/src/serialization/serialization.cpp:115: warning: Member isValidMagicNumber(const char magic[4]) (function) of namespace found is not documented.
/home/tri-nguyen/found/src/serialization/encoding.hpp:84: warning: Member d (variable) of union found::_d_u_ is not documented.
/home/tri-nguyen/found/src/serialization/encoding.hpp:85: warning: Member u (variable) of union found::_d_u_ is not documented.
/home/tri-nguyen/found/src/serialization/encoding.hpp:76: warning: Member f (variable) of union found::_f_u_ is not documented.
/home/tri-nguyen/found/src/serialization/encoding.hpp:77: warning: Member u (variable) of union found::_f_u_ is not documented.
```
. Please update your cpplint version to 2.0.0 and ensure your doxygen is at 1.9.8. Please refer to the `install.sh` script within integration to install these.

src/serialization/serialization.cpp
ozbot11: Right now it is iterating LocationRecords without copying them over.

src/serialization/serialization.cpp
ozbot11: When I initially started, this was a different type. It was just positions. This is leftover.

src/serialization/data_file.cpp
ozbot11: Thanks for the comment, I am being more careful with the checks now.

src/serialization/serialization.cpp
ozbot11: I understand what you mean by unique pointer now. I will add the comment.

src/serialization/data_file.hpp
ozbot11: I made the change.

src/serialization/serialization.cpp
ozbot11: This cannot be done because it changes behavior. They cannot be swapped because the CRC must be calculated on the header after it is converted to network byte order. If you don't convert to network byte order first, the CRC will be computed over the wrong byte representation, causing CRC validation to fail on other systems.

src/serialization/serialization.hpp
nguy8tri, line 1: Please document the precondition that I mentioned in my messages to you do not introduce standardization for `decimal`.

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,153 @@
...
+        throw std::ios_base::failure("Failed to read header");
+    }
+    if (!isValidMagicNumber(header.magic)) {
+        throw std::ios_base::failure("Invalid magic number in header");
+    }
+
+    // Validate CRC
+    uint32_t expected_crc = calculateCRC32(&header, sizeof(header) - sizeof(header.crc));
+    if (ntohl(header.crc) != expected_crc) {
+        std::cerr << "Expected CRC: " << expected_crc << ", Found CRC: " << ntohl(header.crc) << std::endl;
```
nguy8tri, line 143: Replace this with `LOG_ERROR("Expected CRC: " << expected_crc << ", Found CRC: " << ntohl(header.crc));`. You will need to `#include "common/logging.hpp"`

nguy8tri (changes requested):
So, now we have 2 big changes to consider and make, with some minor changes. I give some food for thought in the comments.

test/serialization/serialization-test.cpp
nguy8tri, line 1: I will introduce a mechanism into GitHub Workflow to test this suite further depending on how you choose to handle endianness and `decimal` stuff.

src/serialization/serialization.cpp
nguy8tri: Ah, that makes sense, thanks for clarifying that

src/serialization/encoding.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include <stdint.h>
+
+#include "common/decimal.hpp"
+
+namespace found {
+
+/**
+ * @brief Converts a 16-bit integer from host byte order to network byte order.
+ */
+inline uint16_t htons(uint16_t v) {
```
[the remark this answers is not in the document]
ozbot11: I decided to go with the first solution. I will also update to standardize on double, which may change the code.

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,273 @@
...
+    }
+    value = ntohdec(value);
+}
+
+/**
+ *
+ * @brief Writes a decimal value to the given output stream in network byte order.
+ *
+ */
+inline void write(std::ostream& stream, const decimal& value) {
```
nguy8tri, line 83: If we standardize on `double`, shouldn't we be writing the value as a double?

src/serialization/serialization.cpp
```diff
@@ -0,0 +1,273 @@
...
+    angles.de = ntohdec(angles.de);
+}
+
+/**
+ *
+ * @brief Reads a decimal value from the given input stream.
+ *
+ */
+inline void read(std::istream& stream, decimal& value) {
+    stream.read(reinterpret_cast<char*>(&value), sizeof(decimal));
```
nguy8tri, line 71: If we read a value, shouldn't we be receiving it as a `double`, then using macros to either store it in `value`, or cast it before storing it in `value`?

nguy8tri (changes requested):
Hi Erim, just a last question that I've left, I'm not sure that I'm seeing standardization on `double` as you've specified, and currently, your file will just create a data file where a `decimal` type will change size depending on what used it. Good job encapsulating the read and write methods.

One thing, because they are "private" methods, the documentation won't catch documentation errors with some of the `write` and `read` methods in the `.cpp` methods. However, do document their parameters and return types if you have the time. Thanks!

nguy8tri (approved):

PR #13, opened 2025-05-01
Simple edge detection

sathvikkurap:
# Description
Made the thresholding edge detection algorithm, and I designed two basic tests to validate it (both passing)

# Artifacts for PR #13 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/13/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/13/merge/doxygen)


nguy8tri (changes requested):
Hi Sathvik, after reviewing your code, I have sumarized that there are not only algorithmic changes, there are structural changes, which as members you are not allowed to make unless I've specifically asked for it, or you are very familiar with how to make them (I apologize on that since I haven't been clear on that). As for algorithmic changes, there are also placement and testing issues with your code. Below is a detailed look:

0. Your code fails to build inside the pull request. Usually, you shouldn't make a pull request that fails `make`, and if your PR fails the checks, you are to immediately make changes before requesting another member to review it. It also fails because its not the `Makefile` your code should have (which I will talk about next).
1. In your case, the introduction of `cmake` as well as your modification of `Makefile` are unwarranted. Your specific changes, though they make your code function, would break the functionality of all other existing code, so I am rejecting your modifications to the such. To rectify this, you must revert all changes relavent to the build process (i.e. Delete all CMake files and restore the Makefile to its original state). Conveniently, the current `Makefile` in main is such that it would actually compile your code, and properly. Please also delete all binaries as they also don't belong outside the `build` folder, which is generated by `make`.
2. Your source code looks good, but the location is wrong. It should completely live inside of `src/distance`. More specifically, your code should be part of a class that extends `EdgeDetectionAlgorithm`, which should be defined within `src/distance/edge.hpp` and then implemented in `src/distance/edge.cpp`. Your code specifically belongs to the `Run` method which overrides that base class, and should use `Image`, which is already defined within `style/style.hpp` (which if you notice is different than your definition, and also more efficient). There is no need for a seperate include folder as we do not follow that structure in this repository.
3. Your test suite is in the right place, but your test suite format is wrong. You should be using the `gtest` suite specifications for defining unit tests, an example of which can be found under `test/example/example-test.cpp`. I should also add that your cases are not extensive enough, which will become apparent when a coverage report, which is generated with the original `Makefile`, is produced for your code.

I should add that the majority of these mistakes could've been avoided upon either attending meetings or reading the meeting notes. In meetings we've talked extensively about what the format of the repository is, how it all works together, and how to properly test, and I've often left behind resources for further examination of the such within the notes. Finally, I personally apologize if my way of conveying this information has not been clear, and I ask that you please tell me how if this is the case. But I must also expect members to know this information, as I've taught it through the past year. And finally, thank you for working on this, as this work will be valuable to us and to you!

PR #16, opened 2025-06-17
Enhancement: Iterative Spherical Distance Determination Algorithm (ISDDA)

nguy8tri:
# Description
This PR aims to enhance ISDDA primarily to make it perform better than SDDA. It is in fact capable of doing so.
1. Improved ISDDA to use loss-based selection of the resulting `PositionVector`
- If you want to think of this as a neural network, we (1) use the "network" that `SphericalDistanceDeterminationAlgorithm` defines on randomly selected triplets. We then (2) use a ["Softmax"](https://mriquestions.com/uploads/3/4/5/7/34572113/softmax-example_orig.png) layer on the resulting layer, (3) using a special loss function $\mathcal{L}$ on each input to transform into the logits for each node.
2. Introduced hyperparameters for ISDDA
- A distance ratio defines the ratio of distances between the magnitude of a given `PositionVector` and a reference `PositionVector`, under which 2 different `PositionVector`s will be considered to have the "same distance"
- A loss ratio defines the ratio of losses between the loss for any given `PositionVector` and a reference `PositionVector` (for "step (2)" in SDDA), above which we reject the given `PositionVector` associated that ratio.
- Loss function's radius error power.
- The order of $P$ from below.
3. Use of complex randomization
- Replaced use of `std::random` with more complex seeding, creating triplets with the following method:
   - Uniformly generate the first point, i.e. $i_0 \sim Unif(0, n-1)$ where $n-1$ is the maximum index
   - Non-uniformly generate the second point via $i_1 \sim F(i, i_0)=\frac{P(i, i_0)}{\sum_{i=0}^{n-1}{P(i, i_0)}}$, where $i_1$ is the sampled point and $i$ is any index, and $P(i, i_0)$ is any function such that $P(i_0, i_0)=0$. In our case, we choose $P(i, i_0)=(i-i_0)^2$ (i.e. we are sampling from a discrete quadratic distribution about $i_0$).
   - Non-uniformly generate the third point by sampling via $i_2 \sim \frac{G(i, i_0, i_1)}{\sum_{i=0}^{n-1}{G(i, i_0, i_1)}}$ where the function $G(i, i_0, i_1) = F(i, i_0) + F(i, i_1)$ for $i \neq i_0 \neq i_1$ and $G(i, i_0, i_1) = 0$ otherwise (i.e. we are sampling from a biquadratic function about $i_0$ and $i_1$).
   - Repeat for the rest of the triplets
 - We are not concerned about re-sampling of points between triplets, and resampling doesn't happen within a triplet because
4. Integration into Command-Line
- Hyperparameters are available via the command-line
- You may choose which distance determination algorithm (DDA) to use as well

Testing (In non-optimized mode):
On `test/common/assets/example_earth1.png`:
- SDDA -> (1.0456e+07, -67903.8, -972.935) m
  - Distance Error: 0.752384891562%
  - Angle Error: 1339.6805912772 arcseconds
  - Execution Time: < 1 sec
- ISDDA(100000 iterations, 0.8 distance ratio, INF discriminator, Quadratic $P$ AND $\mathcal{L}$) -> (1.0384e+07, -12571.3, -1057.05) m
   - Distance Error: 0.0565676042517%
   - Angle Error: 250.59497104116 arcseconds
   - Execution Time: 11 sec
- ISDDA(100000 iterations, 0.8 distance ratio, INF discriminator, Quartic $P$ OR $\mathcal{L}$) ->(1.03781e+07, -11536.7, -927.331) m
   - Distance Error: 0.000294332681557%
   - Angle Error: 230.031583013 arcseconds
   - Execution Time: 11 sec
In optimized mode (-O3), all algorithms are less than 1 second.

# Artifacts for PR #16 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/16/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/16/merge/doxygen)


j4lando (approved):
Thanks Tri this looks super cool!

PR #29, opened 2025-07-09
CI/CD: CMake Integration

### LINKED ISSUE #28: CI/CD: Support for CMake
# Description

I would like to add support for `CMake` that does effectively the same thing that `GNU Make` currently does, and bring our build system up to date. Some things that `CMake` offers that `GNU Make` cannot:
1. Ensure versioning of depedency packages or command-line programs
2. Less verbose build processes
3. Easy integration of other CMake projects (e.g. GoogleTest)

The `CMake` build should have a singular `CMakeLists.txt` file that does almost the same thing that the `Makefile` currently does. Some flexibility:
1. You do not need to specify where the `.o` files will go, and just let `CMake` determine that for ease.
2. Any dependency that is a `CMake` project (for instance, GoogleTest) no longer needs to go into the `.cache` and can just be `FetchContent`ed.

Things you cannot compromise:
1. The only file you can make changes to/add is `CMakeLists.txt`. Especially for all configuration files (e.g. `Doxyfile`, `gcovr.cfg`), you cannot change their contents during this process (you may have to run their respective commands from the `CMAKE_SOURCE_DIR`)
2. You can only have 1 `CMakeLists.txt` in the root directory
3. You must have the same targets that do the same things.
4. You must ensure proper versions for anything that you can do for (e.g. libraries like GoogleTest, and command-line programs like `doxygen`)

Finally, `Cmake` has a kind of verbose build process. Thus, you should make a wrapper shell script `build.sh` that accepts the build system and the target:

`./build <cmake OR make> <target>`

This project requires some research into `CMake`. Keep in mind the advantages that `CMake` has and make sure you use them to shorten the length of your build file code. You must also thoroughly understand the current `Makefile`, or you will not really know how to correctly make your `CMake` file. I will say that chatgpt can be useful here, but to a point, as it doesn't understand our `Makefile` completely (I tried this).

nguy8tri:
# Description
This pull request adds `CMake` as a new build system for this project, and updates it to be the default build system. This pull request closes #28.

Changes:
1. New `CMakeLists.txt` which can be used to do everything that `Makefile` currently does.
    a) Instead of `google_stylecheck` and `google_stylecheck_test`, its just one `lint` now
    b) Instead of `doxygen_generate`, its just `documentation`
    c) No setup targets, as those are automatically taken care of
    d) No `private` target`
    d) A cache is still built for our system, but you may consider removing it in the future (will have to be online to compile your tests, or just don't `make clean`) because the alternative is much cleaner.
2. The new build system has additional features:
    a) Ensures correct version for `gcovr`, `cpplint` and `doxygen`
    b) Most external dependencies can be brought in using git URL and commit/tag/branch names
    c) Significantly more readable
    d) Installation is supported through the release target
    e) Readily usable by another CMake project
3. Very minor change to `gcovr.cfg` to exclude all code in the cache.
4. A new bash script `./build` can be used to build the code in both build systems
5. The workflow has now changed to include `CMake`

# Artifacts for PR #29 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/29/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/29/merge/doxygen)

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,3 +140,24 @@ The distance information will then be used to form a vector of the satellite rel
 This stage takes multiple vectors of the satellite at different points in the satellite's orbit to project the satellite's future path of travel. FOUND will be capable of:
 - [ ] Runge-Kutta Based Orbital Prediction
 - [ ] Lambert's Problem-Based Orbital Prediction
+
+# CMake Integration
+
+This project's code can be integrated with yours via cmake. To use:
+```CMake
+include(FetchContent)
+
+cmake_minimum_required(VERSION 3.16)
+project(test)
+
+FetchContent_Declare(
+  found
+  GIT_REPOSITORY https://github.com/UWCubeSat/found.git
+)
+
+FetchContent_MakeAvailable(found)
+
+add_executable(your_executable ...)
+target_link_libraries(test PRIVATE found::found_lib) # Link our library
+```
+The library `found::found_lib` already has all the include statements needed inside of it, so no need to locate them.
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,27 @@
+# FOUND Tools
+
+This folder hosts tools that can be used with FOUND for analysis or artificial generation.
+
+# Usage
+
+Tool Dependencies:
+- Python
+- Highly Recommended: Conda/Miniconda
+
+Each tool uses a different set of dependencies, so its in your best interest to use `conda`. To use each tool:
+1. `cd` into the respective folder
+2. Create a virtual environment based on the `environment.yml` listed (`conda env create -f environment.yml`)
+3. `cd` back into the `found` directory
+4. Activate the environment for that tool (`conda activate <environment-name>`)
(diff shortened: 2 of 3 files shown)


build.sh
```diff
@@ -0,0 +1,22 @@
...
+elif [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
+    echo "Usage: ./build.sh make [GNU Make Options]"
+    echo "Usage: ./build.sh cmake \"[CMake Config Options]\" \"[CMake Build Options]\""
+    echo "Usage: ./build.sh clean"
+    echo "Usage: ./build.sh clean_all"
+elif [ "$1" = "clean" ]; then
+    rm -rf build
+elif [ "$1" = "clean_all" ]; then
+    rm -rf build .cache
+else
```
j4lando, line 19: Could this be `elif [ "$1" = "make" ]; then` to be more precise.

j4lando (approved):

PR #39, opened 2025-09-28
Fix: LOSTVectorGenerationAlgorithm

### LINKED ISSUE #38: Bug: LOSTVectorGenerationAlgorithm
# Issue

There is a glaring bug in this algorithm, but everything passes which is very mysterious. Consider the following code:

```C++

#include <iostream>

#include "common/decimal.hpp"
#include "common/spatial/attitude-utils.hpp"
#include "distance/vectorize.hpp"

using found::Quaternion;
using found::SphericalToQuaternion;
using found::LOSTVectorGenerationAlgorithm;
using found::PositionVector;

int main() {
    // Orientation of FOUND
    Quaternion ori(SphericalToQuaternion(DECIMAL_M_PI_2, 0, 0));

    // Orientation of FOUND relative to LOST (FOUND is 90 degrees behind LOST)
    Quaternion rel(SphericalToQuaternion(3 * DECIMAL_M_PI_2, 0, 0));

    // Orientation of LOST
    Quaternion ref(SphericalToQuaternion(DECIMAL_M_PI, 0, 0));

    // This should obtain the orientation of FOUND
    Quaternion resOri(rel * ref);

    // Running these confirms that ori is equivalent to resOri (they may be negative of each other)
    std::cout << ori.real << " " << ori.i << " " << ori.j << " " << ori.k << std::endl;
    std::cout << resOri.real << " " << resOri.i << " " << resOri.j << " " << resOri.k << std::endl;

    // See comment below
    LOSTVectorGenerationAlgorithm algo1(rel, ref);
    LOSTVectorGenerationAlgorithm algo2(ori);

    PositionVector vec(1, 2, 3);

    PositionVector res1(algo1.Run(vec));
    PositionVector res2(algo2.Run(vec));

    std::cout << std::endl;


    std::cout << res1.x << " " << res1.y << " " << res1.z << std::endl;
    std::cout << res2.x << " " << res2.y << " " << res2.z << std::endl;
}
```

Starting at `LOSTVectorGenerationAlgorithm algo1(rel, ref);`, the first constructor performs `Quaternion resOri(rel * ref)` to obtain the quaternion to ultimately rotate the position vector we obtain from a previous algorithm. However, `LOSTVectorGenerationAlgorithm algo2(ori);` takes the conjugate of `ori` before performing rotation, which at the print statements at the bottom, results in obviously different results. However, the tests in `test/distance/vectorize-test.cpp` and `test/integration/integration-test.cpp` heavily disagree with this result, and claim that they should be the same.

# Diagnosis

The second constructor (which accepts one orientation) is most definitely wrong here. Work backwards from there, fixing all test cases until all test cases comply.

nguy8tri:
# Description
The `LOSTVectorGenerationAlgorithm` is fixed, and successfully closes #38. What we did:
- Identified the root cause (inconsistent conjugation of `orientation` before rotation)
- Updated all test cases to reflect this
- ~Updated the Quaternion math to use forward quaternions instead~

This bug was a result of my negligence and misunderstanding of how the native `Quaternion` interacts with `SphericalToQuaternion`, which produces backwards rotations instead of forward rotations.

To mitigate this issue in the future, documentation has been added to clearly state how that functionality works, with many redundancies.

# Artifacts for PR #39 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/39/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/39/merge/doxygen)


j4lando (commented):
Thank you Tri! This would have been a hard bug to fix since no one is familiar with quaternions. I see the warning about using a forwards quaternion will that change integration with LOST. Does this need to be explained in the README?

j4lando (commented):
To make sure I'm on the same page the constructor with one orientation was correct and the constructor with (rel, rep) was incorrect this is different than the diagnosis of #38

j4lando (approved):

PR #42, opened 2025-10-04
Tool: Attitude Generator

### LINKED ISSUE #19: Bug: Regex Issues
Tests within `test/command-line/execution/executors-test.cpp` and `test/integration/integration-test.cpp` have regex matchers that would not function on computers with different architectures. For instance, the regex used in `TestCalibrationPipelineExecutor` from `test/command-line/execution/executors-test.cpp` uses:

```
expectedOutput << "\\[INFO\\s[0-9]{4}-[0-9]{2}-[0-9]{2}\\s[0-9]{2}:[0-9]{2}:[0-9]{2}\\s[A-Z]+\\] "
                    << "Calibration Quaternion: \\((0.965926, 1.38778e-17, 0.258819, 5.55112e-17)\\)\\s*";
```
The first line is okay, but the second line is not, using hard-coded constants. Please replace this with the appropriate regex that matches any number. 

> [!IMPORTANT]
> Be mindful of the scientific notation.

### LINKED ISSUE #32: CI/CD: Change back to erclu/check-crlf
# Description
In commit 21e9c07, [‎.github/workflows/build.yml](https://github.com/UWCubeSat/found/commit/21e9c076df5f4ebf98a4eb7c0a5cd43449e8c0d4#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721), lines 82-83 were removed in response to the issue reference in [this pull request](https://github.com/erclu/check-crlf/pull/11). When that pull request finishes, please revert those changes (lines 82-83)

You may consider also removing `.gitattributes` if it only contains syntax to ensure LF line endings. You may also choose to keep it as a secondary check.

### LINKED ISSUE #33: CI/CD: Docker Engine Bug
# Description

A bug is present in Docker Engine version 28.3.0 that prevents our `Makefile`'s coverage target from executing properly. We of course cannot do anything about this at the moment, other than to use CMake's coverage target instead. That change was introduced in #30.

When this [Moby Issue](https://github.com/moby/moby/issues/50431) resolves, and [Github Runner](https://github.com/actions/runner/releases) updates to use a new Docker Engine version that resolves that issue, please update [the build workflow](../blob/main/.github/workflows/build.yml) and change the `float` and `coverage` jobs to use the `make [other args] coverage` target.

### LINKED ISSUE #41: Tool: Generate Random Calibration Parameters
# Description

We need to create a tool that uses sound mathematics, independent from our FOUND system, to generate a random orthonormal bases. This can ultimately be used to fill in calibration and distance parameters needed to verify that our system actually functions as promised.

# Requirements

To begin talking about our requirements, we must think about how FOUND's design works:
1. Calibration: Takes in a local and reference orientation (calibration parameters), which are assumed to be locked relative to each other.
2. Distance: Takes in a reference orientation (test parameter), and uses the result from above to infer the local orientation.

We must also account for our image generator:
1. Sometimes, messing with the orientation of our image generator is better to figure out an image that works than being given a random one.

Thus, we come up with the following requirements:
1. The calibration tool should support generating random series of (RA, DE, ROLL) that all have the same transformation relative to each other [Path 1]
2. The calibration tool should support generating a singular set of calibration parameters and a singular test parameter from a desired local orientation (that is the same as a generated/real image) [Path 2]
3. The calibration tool should output the parameters to be ready for input into both the generator tool and FOUND

# Architecture

1. `argparse` to parse arguments
2. `scipy.spatial.transform`: Generate random calibration transformation
3. Branch
    a. Path 1: `scipy.spatial.transform`: Generate random reference bases and use (3) to generate random local bases
    b. Path 2: `scipy.spatial.transform`: Generate reference basis from the inverse of (2) and use (2) again to generate calibration bases
4. Output results in correct format (if the user wants it to go into a text file, they should handle that themselves by piping it)

Critical Flow: 2, 3

# Testing

Because `scipy.spatial.transform` is a correct library, we can use it both in implementation and testing, so long as we understand the results.

nguy8tri:
# Description
This pull request introduces the attitude generator, a tool that generates reference and local attitudes you can use with FOUND's calibration and distance tools. This opens the window for more extensive integration testing!

## Changes
1. Introduced `tools.attitude`, which closes #41
2. Fixed Regex issues with integration tests, which closes #19
3. Fixed `Makefile` issue with Pipeline tests, which closes #33 
    - Caused by inconsistent `g++` and/or environment which does not allow proper merging of the same instance of a templated class across multiple test files (whew).
    - Solved by merging relevant test files together (`sequential-pipeline-test.cpp` + `modifying-pipeline-test.cpp` = `pipeline-test.cpp`)
    - Workflow file (`.github/workflows/build.yml`) has been reverted to reflect that the build systems are now consistent.
5. Added back `erclu/check-crlf@master`, which closes #32
6. Updated all README.md files to reflect changes
7. Minor changes to generator tool (filename bug, which allowed you to name your file `image.png.png`)

## Usage

The generator tool is very simple to use, and does not require any arguments by default. You can read more about it in its README.md, but one nice thing is that in addition to generating the attitudes, it also generates commands (with FOUND and `tools.generator`) for you to use them with!

# Artifacts for PR #42 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/42/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/42/merge/doxygen)


tools/attitude/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+The `attitude` tool generates orientations to use for the calibration and distance modes of FOUND.
+
+Optional Flags:
+0. `--help`, `-h`: Help
+1. `--use-local`: Use a given local orientation to generate a reference orientation for FOUND
+2. `--local-attitude`: The local orientation to use with the flag above (must be used with above flag).
+3. `--calibration-attitude`: The calibration attitude to use (describes transformation from reference to local frame)
+4. `--num-attitude-pairs`: The number of attitude pairs to generate for the distance mode of FOUND
+
```
j4lando, line 18: I found some of these flags a little confusing. It seems like --local-attitude only works in combination with --use-local and only works for the first attitude generated. Should combine the flags into one flag and specify that it only defines the local attitude for the first attitude generated. 

Also maybe a quick blurb or reference to what local, reference, and calibration attitude are

nguy8tri: These are good points, especially on noticing the redundant flag. Adding in those changes now.

tools/attitude/README.md
```diff
@@ -0,0 +1,96 @@
...
+2. `--local-attitude`: A local attitude to use for one of the local/reference attitude pairs for distance. If specified, this will appear in the first pair for use with the distance pipeline.
+3. `--calibration-attitude`: The calibration attitude to use (describes transformation from reference to local frame)
+4. `--num-attitude-pairs`: The number of attitude pairs to generate for the distance mode of FOUND
+
+# Example Usage
+
+From `found` root:
+```bash
+>>> python3 -m tools.attitude --use-local --local-attitude 45 0 0 --num-attitude-pairs 3
+
```
j4lando, line 24: This uses the old flags

j4lando (changes requested):
Update attitude ReadMe

j4lando (approved):

PR #47, opened 2025-10-28
Feature: Implementing Algorithm Superclass

### LINKED ISSUE #23: Integration: Create Algorithm Superclass
# Description

Create the abstract superclass `EdgeFilteringAlgorithm : ModifiyingStage<Input, Output>`, and integrate it into the `DistanceExecutor` workflow.

> [!NOTE]
> This references a change that must be made in https://github.com/UWCubeSat/found/issues/17 (namely modification 4). Make that change if it has not been completed, and notify those people that that change has been made.

> [!TIP]
> Make this change with one implementation of this other algorithms. This will prevent "undefined reference to..." errors during compilation.

# Integration Details

Integrating this class into `DistanceExecutor` is going to be somewhat different than is shown in [this modification guide](https://www.notion.so/huskysat/FOUND-Architecture-Implementation-24b4a4f5c8e081f3bc4bd52df2bacf0e?source=copy_link#24b4a4f5c8e0819f89e0f0dc2a4d2e4e). The reason for this is because the user may request more than 1 type of filter. Instead of following these instructions to a tee:
1. Instead of making a `std::string` flag that asks what filtering algorithm will be used (e.g. `--distance-algo` is a string flag that figures out which distance determination algorithm to use), specify `OPT_ASSIGN` flags that will "enable" each filtering algorithm.
2. In the file where all the edge filtering algorithms are defined, define a typedef for `ModifyingPipeline<Points>`, called `EdgeFilteringAlgorithms` or something that would make sense.
3. Make a provider method (`ProvideEdgeFilteringAlgorithm`) for a `EdgeFilteringAlgorithms` that constructs that modification pipeline and adds all the requested `EdgeFilteringAlgorithm`s to it as stages.
4. Create a constructor argument for a `EdgeFilteringAlgorithms &&` in `DistanceExecutor`. It should `std::move` the class into `DistanceExecutor`.
5. Modify the factory method for `DistanceExecutor` to reflect these changes, calling the provider method `ProvideEdgeFilteringAlgorithm` to provide the filters. Use constructor overloading if no filter is provided (aka define a constructor without `EdgeFilteringAlgorithms`.

MahirEmran:
# Description
Resolves #23 

* Added EdgeFilteringAlgorithm into DistancePipelineExecutor constructor
* Edited DistancePipeline to have 4 items
* Added edge_filters.hpp w/ ProvideEdgeFilteringAlgorithm and ModifyingPipeline<Point>
* Added OPT_ASSIGN flags

# Artifacts for PR #47 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/47/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/47/merge/doxygen)


nguy8tri:
Hi, so you may notice that your coverage is failing for unapparent reasons. A recent bug that was introduced into Linux is the likely culprit, but unfortunately, we don't know what the bug is or how to fix it. However, we do know a way to go around it. I have noticed that you've added some new tests to `pipeline-tests.cpp`. You should try moving them to `nominal-pipeline-test.cpp` to see if it fixes the issue. If that doesn't work, try making all the pipeline tests in one file.

> [!NOTE]
> For reference, there seems to be an issue where instantiating templates across multiple locations makes for uncombined or incomplete coverage data. Thus, the solution to this issue is to only test a particular instantiation in one file. Attempting to test them in both files (as you've done for `DistancePipeline`) will for some reason make gcovr believe that some things are not fully covered.

Also, I highly encourage you to work with mocks instead of instantiating new objects. This may seem odd at first, but it is a very robust way to mock behavior instead of making full on implementations, however small it looks. For instance, all your test stages within `sequential-pipeline` can easily be reformed into:

```C++
EXPECT_CALL(mock, function(args...)).WillOnce(testing::Return(your stuff))
```

As odd as it seems, it is an industry essential in nominal software development, but also just a good way to not worry about making test classes. This whole thing about mocking falls under something called Dependency Injection, used everywhere in the real world. To learn more about it, please look here: [Mocking and Dependency Injection](https://www.notion.so/huskysat/Our-Continuous-Integration-Deployment-CI-CD-24b4a4f5c8e0819a960ce8989ed9e38a?source=copy_link#24b4a4f5c8e081dba77ddb052a10dd64). You should also look at the purpose of the `providers` folder, detailed in [Executor Injection](https://www.notion.so/huskysat/FOUND-Architecture-Implementation-24b4a4f5c8e081f3bc4bd52df2bacf0e?source=copy_link#24b4a4f5c8e0813db5badbcbab9ada20), which is NEVER to be used directly in test cases (only indirectly through integration tests) :).

For reference the two functions that were not called (and why your coverage is failing in Github) are (there is almost no logic to this by the way):
- [found::SequentialPipeline<std::pair<found::EulerAngles, found::EulerAngles>, found::Quaternion, 1ul>::Run(std::pair<found::EulerAngles, found::EulerAngles> const&) (line 222)]
- [found::SequentialPipeline<std::vector<found::LocationRecord, std::allocator<found::LocationRecord> >, std::vector<found::LocationRecord, std::allocator<found::LocationRecord> >, 2ul>::Run(std::vector<found::LocationRecord, std::allocator<found::LocationRecord> > const&) (line 222)]

src/command-line/execution/executors.cpp
```diff
@@ -53,6 +53,27 @@ DistancePipelineExecutor::DistancePipelineExecutor(DistanceOptions &&options,
                    .Complete(*this->vectorizationAlgorithm);
 }
 
+/**
+ * New overload: accepts an edge-filtering pipeline (EdgeFilteringAlgorithms)
+ * which is moved into the executor and inserted between
+ * edge-detection and distance-determination stages.
+ */
+DistancePipelineExecutor::DistancePipelineExecutor(DistanceOptions &&options,
```
nguy8tri, line 59: Good job on the overload, however:
1.  I'd like to see the documentation go into the `.hpp` file to keep the `.cpp` file clean.
2. Reorder the `filter` to be after edgeDetectionAlgorithm
3. It should be a unique_ptr
4. The algorithm class should have a field for this pointer, just like the other algorithms

src/command-line/execution/executors.hpp
```diff
@@ -75,36 +76,39 @@ class DistancePipelineExecutor : public PipelineExecutor {
...
+                                      EdgeFilteringAlgorithms &&filters);
+
     void ExecutePipeline() override;
     void OutputResults() override;
 
  private:
     /// The DistanceOptions being used
     const DistanceOptions options_;
     /// The Distance pipeline being used
+    // Increase capacity to 4 stages: edge detection -> filters -> distance -> vectorize
```
nguy8tri: Remove comment, since this is understood after above changes.

src/command-line/execution/executors.hpp
```diff
@@ -75,36 +76,39 @@ class DistancePipelineExecutor : public PipelineExecutor {
...
     DistancePipeline pipeline_;
     /// The Edge Detection Algorithm used
     std::unique_ptr<EdgeDetectionAlgorithm> edgeDetectionAlgorithm;
     /// The Distance Determination Algorithm being used
     std::unique_ptr<DistanceDeterminationAlgorithm> distanceAlgorithm;
     /// The Vectorization/Rotation Algorithm being used
     std::unique_ptr<VectorGenerationAlgorithm> vectorizationAlgorithm;
+    /// Optional edge-filtering pipeline (may be empty)
+    EdgeFilteringAlgorithms filters_;
```
nguy8tri: Make `std::unique_ptr` and reorder

src/command-line/parsing/options.hpp
```diff
@@ -64,6 +64,8 @@ FOUND_CLI_OPTION("isdda-discrim-ratio"     , decimal           , ISDDADiscimRati
 FOUND_CLI_OPTION("isdda-pdf-order"         , int               , ISDDAPdfOrd     , 2                          , atoi(optarg)               , kNoDefaultArgument, REQ_ASSIGN, "The Probability Density Function Order for ISSDA (even int)"  ) \
 FOUND_CLI_OPTION("isdda-radius-loss-order" , int               , ISDDARadLossOrd , 4                          , atoi(optarg)               , kNoDefaultArgument, REQ_ASSIGN, "The Radius Loss Order ISSDA (even int)"                       ) \
 FOUND_CLI_OPTION("output-file"             , std::string       , outputFile      , ""                         , optarg                     , kNoDefaultArgument, REQ_ASSIGN, "The output file (*.found)"                                    ) \
+FOUND_CLI_OPTION("enable-noop-edge-filter" , bool              , enableNoOpEdgeFilter, false                  , found::strtobool(optarg)   , true              , OPT_ASSIGN, "Enable the NoOp edge filter (test/demo only)"                 ) \
```
nguy8tri, line 69: Add a note to remove this upon implementation of the first edge filtering algorithm. Since we do not use NoOp Algorithms, we should not encourage this pattern.

src/common/pipeline/pipelines.hpp
```diff
@@ -312,7 +312,9 @@ class ModifyingPipeline : public Pipeline<T, T, N> {
         }
         *this->product = input;
         for (size_t i = 0; i < this->size; i++) {
-            dynamic_cast<ModifyingStage<T> *>(this->stages[i])->SetResource(*this->product);
+            auto *stage = static_cast<ModifyingStage<T> *>(this->stages[i]);
+            assert(stage != nullptr);
```
nguy8tri: This check, although admirable, is impractical in nominal use cases, please get rid of the assertion and go straight for the `SetResource` operation (also meaning you don't need to store a variable)

I'm also interested to hear why you chose a `static_cast` instead of `dynamic_cast`. I think it could be justified, just as long as we also agree on it.

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,79 @@
...
+    EdgeFilteringAlgorithm() = default;
+    virtual ~EdgeFilteringAlgorithm() = default;
+    // Implementations override void Run(Points &)
+};
+
+/**
+ * A trivial implementation to ensure linkage / prevent undefined references.
+ * Does nothing to the Points (no-op filter).
+ */
+class NoOpEdgeFilter : public EdgeFilteringAlgorithm {
```
nguy8tri: Add a note to remove this upon implementation of the first edge filtering algorithm. Since we do not use NoOp Algorithms, we should not encourage this pattern.

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,79 @@
...
+    NoOpEdgeFilter() = default;
+    ~NoOpEdgeFilter() override = default;
+
+    void Run(Points &pts) override {
+        // intentionally no-op
+        (void)pts;
+    }
+};
+
+namespace detail {
```
nguy8tri: 1. We should not be defining single member namespaces, provided stages should be constructed within the provider function

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,79 @@
...
+inline NoOpEdgeFilter kNoOpEdgeFilter{};
+}  // namespace detail
+
```
nguy8tri: Neither of the ProvideEdgeFilterAlgorithm functions should exist here. They should be in `stage-providers`.

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Convenience provider (no-arg). Returns a pipeline that contains a single
+ * NoOpEdgeFilter so callers that expect a ready pipeline can use this.
+ */
+inline EdgeFilteringAlgorithms ProvideEdgeFilteringAlgorithm() {
```
nguy8tri: Further comment here, there never should be a default `ProvideEdgeFilteringAlgorithm`. This is because we would never use this.

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Options-aware provider. Constructs a pipeline containing all filters
+ * that are enabled in options and returns it. If no filter is enabled,
+ * returns std::nullopt to indicate "no filters".
+ *
+ * Note: stages added are function-local static instances so the pipeline
+ * can safely store raw pointers to them.
+ */
+inline std::optional<EdgeFilteringAlgorithms> ProvideEdgeFilteringAlgorithm(const DistanceOptions &options) {
```
nguy8tri: Return type should be unique_ptr, not `std::optional`. If you think about it, it should never be optional, because our executor constructor doesn't accept an optional object. That's why we overload the constructor, because we do know at runtime which one to call and whether it the EdgeFilteringAlgorithms will exist.

src/providers/factory.hpp
```diff
@@ -3,9 +3,11 @@
 
 #include <memory>
 #include <utility>
+#include <optional>
 
 #include "command-line/execution/executors.hpp"
 #include "providers/stage-providers.hpp"
+#include "distance/edge-filters.hpp"  // ProvideEdgeFilteringAlgorithm
```
nguy8tri: No need for a comment here

src/providers/factory.hpp
```diff
@@ -29,10 +31,32 @@ inline std::unique_ptr<CalibrationPipelineExecutor> CreateCalibrationPipelineExe
  * @return A pointer to a DistancePipelineExecutor
  */
 inline std::unique_ptr<DistancePipelineExecutor> CreateDistancePipelineExecutor(DistanceOptions &&options) {
-    return std::make_unique<DistancePipelineExecutor>(std::forward<DistanceOptions>(options),
-                                    ProvideEdgeDetectionAlgorithm(std::forward<DistanceOptions>(options)),
-                                    ProvideDistanceDeterminationAlgorithm(std::forward<DistanceOptions>(options)),
-                                    ProvideVectorGenerationAlgorithm(std::forward<DistanceOptions>(options)));
+    // Read enable flags before anything that might move resources.
+    bool enableNoop = options.enableNoOpEdgeFilter;
```
nguy8tri: For implementing just the algorithm superclass, the no-op algorithm is necessary for the construction of a modifying pipeline. However, based on the options we know if a `EdgeFilteringAlgorithms` will exist or not. Thus, your code should reflect this without resorting to `std::optional` (which I know I addressed above).

In my opinion, a singular provider for `EdgeFilteringAlgorithms` should exist (that will later figure out which algorithms to use and in what order)

src/providers/stage-providers.hpp
```diff
@@ -53,19 +53,14 @@ std::unique_ptr<EdgeDetectionAlgorithm> ProvideEdgeDetectionAlgorithm(DistanceOp
  * 
  * @return std::unique_ptr<DistanceDeterminationAlgorithm> The distance determination algorithm
  */
-std::unique_ptr<DistanceDeterminationAlgorithm> ProvideDistanceDeterminationAlgorithm(DistanceOptions &&options) {
+inline std::unique_ptr<DistanceDeterminationAlgorithm> ProvideDistanceDeterminationAlgorithm(const DistanceOptions &options) {
     if (options.distanceAlgo == SDDA) {
-        return std::make_unique<SphericalDistanceDeterminationAlgorithm>(options.radius,
```
nguy8tri, line 58: There's actually a point to constructing it this way. Construction directly in the constructor allows the function to save memory, instead of constructing it in the function and then copying it.

C++ is funny like that, it doesn't optimize things well sometimes unless you tell it too.

test/command-line/execution/executors-test.cpp
```diff
@@ -83,26 +83,26 @@ TEST(ExecutorsTest, TestCalibrationPipelineExecutor) {
...
-        "hello",
-        92,
-        300,
-        2.0,
-        10,
-        12,
-        temp_df
-    };
+    DistanceOptions options;
+    options.image = strtoimage("test/common/assets/example_image.jpg");
```
nguy8tri, line 85: This is great, now we don't have to guess what is what haha

test/command-line/execution/executors-test.cpp
```diff
@@ -134,10 +134,15 @@ TEST(ExecutorsTest, TestDistancePipelineExecutor) {
     std::unique_ptr<VectorGenerationAlgorithm>
         vectorGenerationAlgorithm(std::move(mockVectorGenerationAlgorithm));
 
+    // using new edge-filtering pipeline integration.
+    // ProvideEdgeFilteringAlgorithm returns an EdgeFilteringAlgorithms (ModifyingPipeline<Points>)
+    // that currently contains a NoOpEdgeFilter so behavior is unchanged.
+    auto filters = found::ProvideEdgeFilteringAlgorithm();
```
nguy8tri: No, mock this, as in just like the other pipeline executors, make a mock for the filtering "stage". One of the tenants of Dependency Injection is that even if ProvideEdgeFilteringAlgorithm changes, our test case will not. We fundamentally don't care what the "stage" is, just that it is called within the executor, which we can ensure with mocking.

test/command-line/execution/executors-test.cpp
```diff
@@ -170,6 +175,94 @@ TEST(ExecutorsTest, TestDistancePipelineExecutor) {
     std::remove(temp_df);
 }
 
+// new test using the no-filter constructor
+TEST(ExecutorsTest, TestDistancePipelineExecutor_NoFilters) {
```
nguy8tri: I dislike the use of `_` within test case functions, please don't use underscores, just do capitalized naming conventions like the other test cases.

test/command-line/parsing/parser-test.cpp
```diff
@@ -219,4 +219,37 @@ TEST_F(ParserTest, OrbitParserGeneral) {
     ASSERT_DECIMAL_EQ_DEFAULT(DECIMAL(400000.0), options.mu);
 }
```
nguy8tri, line 223: Great job on following the instructions for OPT_ASSIGN coverage! That can be a bit confusing I'll admit!

test/common-test/pipeline/pipeline-test.cpp
```diff
@@ -631,4 +662,90 @@ TEST(ModifyingPipelineTest, TestSequentialAndModifierPipelinesinSequentialPipeli
     ASSERT_EQ(expectedResource, resource);
 }
 
+// Alias-based smoke test for DistancePipeline to ensure the typedef stays wired
```
nguy8tri: I don't know about you, but I personally don't see value in these test cases, and I'll explain why.

I believe what you mean by this comment (I am guessing, so correct me if I'm wrong) is the change from 3 to 4 for `DistancePipeline`. The only concern in this case is if you were to put more than the capacity in `DistancePipeline`, which it cannot handle. Luckily, one of the features of the test cases is that it uses both `asan` and `valgrind`, which check for improper memory usage such as that case I just mentioned (called buffer overflow). Because this check happens automatically, and because our code is 100% covered, we're not concerned about that, so I think you should be able to safely delete these.

test/common-test/pipeline/pipeline-test.cpp
```diff
@@ -631,4 +662,90 @@ TEST(ModifyingPipelineTest, TestSequentialAndModifierPipelinesinSequentialPipeli
...
+    PointsToVec3Stage stage2;
+
+    pipeline.AddStage(stage1).Complete(stage2);
+
+    Image input{};
+    Vec3 output{};
+    pipeline.GetResource() = input;
+    pipeline.GetProduct() = &output;
+
+    pipeline.DoAction(); // This directly calls the target function.
```
nguy8tri: This is unnecessary, this is already invoked within `nominal-pipeline-test.cpp`

test/common-test/pipeline/pipeline-test.cpp
```diff
@@ -631,4 +662,90 @@ TEST(ModifyingPipelineTest, TestSequentialAndModifierPipelinesinSequentialPipeli
...
+
+    ASSERT_EQ(output.x, 5);
+    ASSERT_EQ(output.y, 6);
+    ASSERT_EQ(output.z, 7);
+}
+
+// Directly exercise FunctionStage<Points, Points>::DoAction
+TEST(FunctionStageTest, DoActionPointsIdentityStage) {
```
nguy8tri: I think you won't need this when you mock a FunctionStage<Points, Points> within the executors test

test/common-test/pipeline/pipeline-test.cpp
```diff
@@ -631,4 +662,90 @@ TEST(ModifyingPipelineTest, TestSequentialAndModifierPipelinesinSequentialPipeli
...
+
+    ASSERT_EQ(product.size(), input.size());
+    for (size_t i = 0; i < product.size(); ++i) {
+        EXPECT_EQ(product[i].x, input[i].x + DECIMAL(1.0));
+        EXPECT_EQ(product[i].y, input[i].y + DECIMAL(2.0));
+    }
+}
+
+// Explicit coverage for SequentialPipeline<Image, Vec3, 4>::Run via a member pointer call
+TEST(SequentialPipelineTest, RunImageToVec3N4CoversRun) {
```
nguy8tri: I'm more confused why this is necessary here? I'm not convinced this is really doing anything for coverage.

test/common-test/pipeline/pipeline-test.cpp
```diff
@@ -631,4 +662,90 @@ TEST(ModifyingPipelineTest, TestSequentialAndModifierPipelinesinSequentialPipeli
...
+    Image input{};
+    auto runPtr = &SequentialPipeline<Image, Vec3, 4>::Run;
+    Vec3 result = (pipeline.*runPtr)(input);
+
+    ASSERT_EQ(result.x, 5);
+    ASSERT_EQ(result.y, 6);
+    ASSERT_EQ(result.z, 7);
+}
+
+// Also verify the alias instantiation uses the same Run logic
```
nguy8tri: Alias instantiation doesn't affect coverage. Why? Because all aliases are recursively resolved into their original C++ types within gcovr (as well as the compiler to some extent)!

test/distance/edge-filters-test.cpp
```diff
@@ -0,0 +1,115 @@
+#include <gtest/gtest.h>
+
+#include "test/common/common.hpp"
+#include "src/distance/edge.hpp"
+#include "src/distance/edge-filters.hpp"
+#include "src/providers/factory.hpp"
+
+namespace found {
+
```
nguy8tri, line 6: As a general note on test conventions

1. Don't use underscores in the test names
2. Only 1 new line between test cases

test/distance/edge-filters-test.cpp
```diff
@@ -0,0 +1,115 @@
+#include <gtest/gtest.h>
+
+#include "test/common/common.hpp"
+#include "src/distance/edge.hpp"
```
nguy8tri: In test cases, as a convention, please try to follow the conventional grouping of import statements, seperated by spaces:
1. External C Libraries
2. External C++ Libraries
3. Our Libraries
4. Our Test Libraries

You'll also find this convention mentioned [here](https://www.notion.so/huskysat/C-Software-Style-24b4a4f5c8e081abaa17da8bf58cc52b?source=copy_link#24b4a4f5c8e0819796b5ee2f2b876c51)

test/distance/edge-filters-test.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "src/distance/edge-filters.hpp"
+#include "src/providers/factory.hpp"
+
+namespace found {
+
+TEST(EdgeFiltersTest, EdgeFilteringAlgorithms_RunWithoutCompletionThrows) {
```
nguy8tri: This is already guarenteed per the `pipeline` tests. No need for duplicate tests

test/distance/edge-filters-test.cpp
```diff
@@ -0,0 +1,115 @@
...
+    EXPECT_THROW(filters.Run(pts), std::runtime_error);
+}
+
+TEST(EdgeFiltersTest, ProvideEdgeFilteringAlgorithm_NoOpDoesNothing) {
+    Points original = {
+        {0, 0},
+        {1, 1},
+        {2, 3}
+    };
+    auto filters = ProvideEdgeFilteringAlgorithm();
```
nguy8tri: I think I've seen `auto` in many places at this point, however, you are using it for convenience more than readability, and it is actually somewhat confusing to me what certain things should be. As you will find in this comment [here](https://www.notion.so/huskysat/C-Software-Style-24b4a4f5c8e081abaa17da8bf58cc52b?source=copy_link#24b4a4f5c8e0813ebe39f81ef64334b5), the Google C++ Style Guide recommends limitation of the keyword `auto` unless you get readability benefits.

In our case, this means that unless the name of the variable is long (e.g. A type qualified by many layers of namespaces, or a Type with many type parameters), you should not use `auto`.

test/distance/edge-filters-test.cpp
```diff
@@ -0,0 +1,115 @@
...
+    Points result = filters.Run(original);
+    ASSERT_EQ(original.size(), result.size());
+    for (size_t i = 0; i < original.size(); ++i) {
+        ASSERT_EQ(original[i].x, result[i].x);
+        ASSERT_EQ(original[i].y, result[i].y);
+    }
+}
+
+
+TEST(EdgeFiltersTest, ProvideEdgeFilteringAlgorithm_OptionsDisabledReturnsNullopt) {
```
nguy8tri: No direct invocations of `Provide*` or `Create*` in your test cases. These are Dependency Injection functions, and should indirectly invoked via the integration tests.

nguy8tri (changes requested):
I was going to leave this to Josh, but I realized some confusion and tough issues warrant my input on this.

I want to first commend you on doing all this, its not easy to pick up a new code space and to run with it. That being said, there are many things that must be corrected. I think you will see that a lot of this could've been avoided by looking over the thourough documentation in FOUND (but who does that am I right haha?).

A summary of what I saw:
1. I saw many opportunities for Dependency Mocking. Trust me, its worth your time, so take some time to learn how its used via the links in the comments.
2. I also saw a lack of understanding around Dependency Injection Framework that's used in the code. No worries, that's not even taught in the 4 years (in my case 5) that you're at UW.
3. Confusion around the coverage. That's understandable, we'll work through it together (for reference, we faced the same issue before, and it took me 3 months to figure it out)
4. Unnecessary code that is good in theory but useless/unoptimized in production. That will take a while to shake itself out (I'm still doing that myself)
5. Nitpicky style things to help you conform to the rest of the code.
6. (Were you using AI? No pressure, I'm not your professor, and I use it a lot in my work as a software dev. However, some things seemed out of place so I thought I'd ask for curiosity sake.)

Anyways, I hope the comments here will be helpful! Please ask me any questions you might have!

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <optional>
+
+#include "command-line/parsing/options.hpp"
+#include "common/pipeline/pipelines.hpp"
+#include "distance/edge.hpp"
+
+namespace found {
+
+
+using EdgeFilteringAlgorithms = ModifyingPipeline<Points>;
```
nguy8tri: I much prefer typedef here, since it was made for this (so `typedef ModifyingPipeline<Points> EdgeFilteringAlgorithms`).

src/common/pipeline/pipelines.hpp
MahirEmran: I believe I did this mainly just because of testing in pipelines and just put an assert there to make sure it was fine. Forgot to change  it back later (I thought it may have been an issue with polymorphism or such).

src/command-line/execution/executors.cpp
```diff
@@ -53,6 +53,23 @@ DistancePipelineExecutor::DistancePipelineExecutor(DistanceOptions &&options,
...
+                                                   std::unique_ptr<EdgeDetectionAlgorithm> edgeDetectionAlgorithm,
+                                                   std::unique_ptr<EdgeFilteringAlgorithms> filters,
+                                                   std::unique_ptr<DistanceDeterminationAlgorithm> distanceAlgorithm,
+                                                   std::unique_ptr<VectorGenerationAlgorithm> vectorizationAlgorithm)
+                                                   : options_(std::move(options)), filters_(std::move(filters)) {
+    this->edgeDetectionAlgorithm = std::move(edgeDetectionAlgorithm);
+    this->distanceAlgorithm = std::move(distanceAlgorithm);
+    this->vectorizationAlgorithm = std::move(vectorizationAlgorithm);
+    // Build the pipeline: edge detection -> filters (modifying Points) -> distance -> vectorize
```
nguy8tri: Get rid of comment and put in proper order

src/command-line/execution/executors.hpp
```diff
@@ -75,19 +76,30 @@ class DistancePipelineExecutor : public PipelineExecutor {
...
-     * 
-     * @pre options.image.image must be point to heap allocated memory.
-     * This is guarenteed as long as strtoimage is used to create the image,
-     * and it throws an error if the image is not valid.
+     * Constructs a DistancePipelineExecutor (no edge-filters)
+     *
+     * @param options The DistanceOptions to configure the pipeline (moved into the executor)
+     * @param edgeDetectionAlgorithm The EdgeDetectionAlgorithm used by the pipeline (moved into the executor)
+     * @param distanceAlgorithm The DistanceDeterminationAlgorithm used by the pipeline (moved into the executor)
+     * @param vectorizationAlgorithm The VectorGenerationAlgorithm used by the pipeline (moved into the executor)
```
nguy8tri, line 82: Where did the preconditions go?

src/command-line/execution/executors.hpp
```diff
@@ -75,19 +76,30 @@ class DistancePipelineExecutor : public PipelineExecutor {
...
+
+    /**
+     * Constructs a DistancePipelineExecutor with an edge-filtering pipeline
+     *
+     * @param options The DistanceOptions to configure the pipeline (moved into the executor)
+     * @param edgeDetectionAlgorithm The EdgeDetectionAlgorithm used by the pipeline (moved into the executor)
+     * @param filters A pipeline of edge filtering stages; ownership is transferred to the executor
+     * @param distanceAlgorithm The DistanceDeterminationAlgorithm used by the pipeline (moved into the executor)
+     * @param vectorizationAlgorithm The VectorGenerationAlgorithm used by the pipeline (moved into the executor)
      */
```
nguy8tri, line 105: Same note, where are my preconditions

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Abstract superclass for edge-filtering algorithms.
+ * modifies Points in-place (ModifyingStage<Points>).
+ */
+class EdgeFilteringAlgorithm : public ModifyingStage<Points> {
+ public:
+    EdgeFilteringAlgorithm() = default;
+    virtual ~EdgeFilteringAlgorithm() = default;
+    // Implementations override void Run(Points &)
```
nguy8tri: This is understood remove comment

src/providers/factory.hpp
```diff
@@ -29,10 +31,23 @@ inline std::unique_ptr<CalibrationPipelineExecutor> CreateCalibrationPipelineExe
...
  */
 inline std::unique_ptr<DistancePipelineExecutor> CreateDistancePipelineExecutor(DistanceOptions &&options) {
-    return std::make_unique<DistancePipelineExecutor>(std::forward<DistanceOptions>(options),
-                                    ProvideEdgeDetectionAlgorithm(std::forward<DistanceOptions>(options)),
-                                    ProvideDistanceDeterminationAlgorithm(std::forward<DistanceOptions>(options)),
-                                    ProvideVectorGenerationAlgorithm(std::forward<DistanceOptions>(options)));
+    std::unique_ptr<EdgeDetectionAlgorithm> edgeAlg = ProvideEdgeDetectionAlgorithm(options);
+    std::unique_ptr<DistanceDeterminationAlgorithm> distAlg = ProvideDistanceDeterminationAlgorithm(options);
+    std::unique_ptr<VectorGenerationAlgorithm> vecAlg = ProvideVectorGenerationAlgorithm(options);
+    std::unique_ptr<EdgeFilteringAlgorithms> filtersOpt = ProvideEdgeFilteringAlgorithm(options);
```
nguy8tri: Reorder

src/providers/stage-providers.hpp
```diff
@@ -40,7 +42,7 @@ std::unique_ptr<CalibrationAlgorithm> ProvideCalibrationAlgorithm([[maybe_unused
  * 
  * @return std::unique_ptr<EdgeDetectionAlgorithm> The edge detection algorithm
  */
-std::unique_ptr<EdgeDetectionAlgorithm> ProvideEdgeDetectionAlgorithm(DistanceOptions &&options) {
+inline std::unique_ptr<EdgeDetectionAlgorithm> ProvideEdgeDetectionAlgorithm(const DistanceOptions &options) {
```
nguy8tri: Should be r value reference

src/providers/stage-providers.hpp
```diff
@@ -53,19 +55,17 @@ std::unique_ptr<EdgeDetectionAlgorithm> ProvideEdgeDetectionAlgorithm(DistanceOp
  * 
  * @return std::unique_ptr<DistanceDeterminationAlgorithm> The distance determination algorithm
  */
-std::unique_ptr<DistanceDeterminationAlgorithm> ProvideDistanceDeterminationAlgorithm(DistanceOptions &&options) {
+inline std::unique_ptr<DistanceDeterminationAlgorithm> ProvideDistanceDeterminationAlgorithm(
```
nguy8tri, line 60: Same here

src/providers/stage-providers.hpp
```diff
@@ -84,7 +84,7 @@ std::unique_ptr<DistanceDeterminationAlgorithm> ProvideDistanceDeterminationAlgo
  * 
  * @return std::unique_ptr<VectorGenerationAlgorithm> The vector generation algorithm
  */
-std::unique_ptr<VectorGenerationAlgorithm> ProvideVectorGenerationAlgorithm(DistanceOptions &&options) {
+inline std::unique_ptr<VectorGenerationAlgorithm> ProvideVectorGenerationAlgorithm(const DistanceOptions &options) {
```
nguy8tri: Must be r value reference

src/providers/stage-providers.hpp
```diff
@@ -100,6 +100,47 @@ std::unique_ptr<VectorGenerationAlgorithm> ProvideVectorGenerationAlgorithm(Dist
...
 
+/**
+ * NoOpEdgeFilter
+ *
+ * A ModifyingStage implementation that performs no modifications
+ * to the Points. This exists to provide a valid stage instance that can be
+ * used by providers when a no-op filter is requested. This will not be needed
+ * once an EdgeFilteringAlgorithm is implemented.
+ */
+class NoOpEdgeFilter : public ModifyingStage<Points> {
```
nguy8tri: This is in the wrong file, please put this back at the edge-filter file

src/providers/stage-providers.hpp
```diff
@@ -100,6 +100,47 @@ std::unique_ptr<VectorGenerationAlgorithm> ProvideVectorGenerationAlgorithm(Dist
...
+    }
+};
+
+/**
+ * detail
+ *
+ * Internal provider instances and singletons used by the stage provider
+ * functions
+ */
+namespace detail {
```
nguy8tri: I should've been explicit here, but unless there is a point, never make another namespace

src/providers/stage-providers.hpp
```diff
@@ -100,6 +100,47 @@ std::unique_ptr<VectorGenerationAlgorithm> ProvideVectorGenerationAlgorithm(Dist
...
+/**
+ * A single shared instance of the NoOpEdgeFilter used by providers when the
+ * no-op edge filter is requested. Documented to satisfy Doxygen checks.
+ */
+inline NoOpEdgeFilter kNoOpEdgeFilter{};
```
nguy8tri: Should be constructed in provider function

test/common/mocks/distance-mocks.hpp
```diff
@@ -24,6 +24,20 @@ class MockVectorGenerationAlgorithm : public VectorGenerationAlgorithm {
     MOCK_METHOD(PositionVector, Run, (const PositionVector& points), (override));
 };
 
+class MockEdgeFilter : public ModifyingStage<Points> {
```
nguy8tri: Redundant class

test/distance/edge-filters-test.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include "src/distance/edge.hpp"
+#include "src/distance/edge-filters.hpp"
+
+#include "test/common/mocks/distance-mocks.hpp"
+#include "test/common/common.hpp"
+
+namespace found {
+
+TEST(EdgeFiltersTest, RunOnEmptyThrows) {
```
nguy8tri: Revaluate which test need to be here. Are we testing the behavior of the pipleine/stage or are we testing the behavior of our actual algorithm. Try removing these one by one to see which ones you need.

nguy8tri (changes requested):
Good job on the changes. However there were a few changes thay weren't done, so I've explicitly marked them. Let me know if you have questions

MahirEmran:
I had to make a merge in executors-test.cpp because (maybe some commit between last commit and now) there was difference in how DistanceOptions obj was being constructed. I feel like the one that was there before was odd (set all the fields in constructor rather than initializing object, then explicitly setting fields after), so I just went with current changes.

Had to rerun documentation check because first check failed the juliandate time test both attempts.

src/common/pipeline/pipelines.hpp
```diff
@@ -62,6 +65,8 @@ class Pipeline : public FunctionStage<Input, Output> {
  protected:
     /// The stages of this
     Action *stages[N];
+    /// Ownership storage for the stages
+    std::unique_ptr<Action> ownedStages[N];
```
nguy8tri: Would it be easier to just get rid of `Action *stages[N];` and just have `std::unique_ptr<Action> ownedStages[N]`

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,74 @@
...
+        (void)pts;
+    }
+};
+
+/**
+ * Options-aware provider. Constructs a pipeline containing all filters
+ * that are enabled in options and returns it. If no filter is enabled,
+ * returns nullptr to indicate "no filters".
+ */
+inline std::unique_ptr<EdgeFilteringAlgorithms> ProvideEdgeFilteringAlgorithm(const DistanceOptions &options) {
```
nguy8tri: Should be under stage-providers

nguy8tri (changes requested):
Great work (and sorry I didn't respond soon). I just requested a change to eliminate a redundant variable and another comment (I'm a bit tired writing this so I don't have more meaningful comments right now). Please resolve all comments that are no longer relavent.

MahirEmran:
I'm not sure how to see the changes that were requested, maybe issue on my end but Github just isn't showing me anything. Maybe I made the edits implicitly? Not sure. 
Resolved comments resolved through previous commits, left the ones about rvalue unresolved and moving the provider to stage-providers since this commit should hopefully resolve them.

src/common/pipeline/pipelines.hpp
```diff
@@ -63,8 +63,6 @@ class Pipeline : public FunctionStage<Input, Output> {
     virtual Output Run(const Input &input) = 0;
 
  protected:
-    /// The stages of this
-    Action *stages[N];
     /// Ownership storage for the stages
     std::unique_ptr<Action> ownedStages[N];
```
nguy8tri: just change this to stages instead of ownedStages

src/command-line/execution/executors.hpp
```diff
@@ -75,19 +74,38 @@ class DistancePipelineExecutor : public PipelineExecutor {
...
-     * @param edgeDetectionAlgorithm The edge detection algorithm to use
-     * @param distanceAlgorithm The distance determination algorithm to use
-     * @param vectorizationAlgorithm The vectorization algorithm to use
-     * 
-     * @pre options.image.image must be point to heap allocated memory.
-     * This is guarenteed as long as strtoimage is used to create the image,
-     * and it throws an error if the image is not valid.
+     * Constructs a DistancePipelineExecutor (no edge-filters)
+     *
+     * @param options The DistanceOptions to configure the pipeline (moved into the executor)
```
nguy8tri: no need for this comment (moved into the executor), we know this from reading the code, and its also an editorial change that no one would have context on years down the line.

src/common/pipeline/pipelines.hpp
```diff
@@ -60,8 +63,8 @@ class Pipeline : public FunctionStage<Input, Output> {
     virtual Output Run(const Input &input) = 0;
 
  protected:
-    /// The stages of this
-    Action *stages[N];
+    /// Ownership storage for the stages
+    std::unique_ptr<Action> ownedStages[N];
```
nguy8tri: call this stages

src/common/pipeline/pipelines.hpp
```diff
@@ -70,39 +73,42 @@ class Pipeline : public FunctionStage<Input, Output> {
...
     /**
-     * Adds a stage to this pipeline
-     * 
-     * @param stage The stage to add to the pipeline
-     * 
-     * @throw std::invalid_argument iff this is already completed
-     * 
+     * Adds a stage to this pipeline, taking ownership of the provided stage.
+     *
+     * @param stage The stage to add to the pipeline (ownership transferred via unique_ptr)
```
nguy8tri: You are stating a postcondition here. For this, you will use the `@post` annotation instead of appending this parenthetical here.

src/common/pipeline/pipelines.hpp
```diff
@@ -70,39 +73,42 @@ class Pipeline : public FunctionStage<Input, Output> {
...
+     *
+     * @throw std::invalid_argument iff this pipeline has already been completed
```
nguy8tri: we use `this` to mean the class itself when we're not in the initial description (Top description of this multi line comment). Its odd, I know, but I want you to get used to the practice. After all, classes represent objects, so referring to ourselves as `this` makes sense.

src/common/pipeline/pipelines.hpp
```diff
@@ -70,39 +73,42 @@ class Pipeline : public FunctionStage<Input, Output> {
...
+     *
      * @pre This method is called when the number of registered stages is
-     * less than N - 1
-     * 
+     *       less than N - 1
```
nguy8tri: one less space

src/common/pipeline/pipelines.hpp
```diff
@@ -70,39 +73,42 @@ class Pipeline : public FunctionStage<Input, Output> {
...
     }
 
     /**
-     * Completes a pipeline with a stage
-     * 
-     * @param stage The stage to add
-     * 
-     * @throw std::invalid_argument iff this is already completed
-     * 
+     * Completes a pipeline with a final stage, taking ownership of it.
```
nguy8tri, line 95: Same comments as with the other javadoc comment

src/common/pipeline/pipelines.hpp
```diff
@@ -162,20 +168,25 @@ class SequentialPipeline : public Pipeline<Input, Output, N> {
      * @pre This method is called when the number of registered stages is
      * less than N - 1
      */
-    template<typename I, typename O> SequentialPipeline &AddStage(FunctionStage<I, O> &stage) {
+    template<typename StageType>
+    SequentialPipeline &AddStage(std::unique_ptr<StageType> stage) {
```
nguy8tri: Wait this makes no sense. Why would you abandon the built in typing of unique_ptr<FunctionStage<I, O>>?

src/common/pipeline/pipelines.hpp
```diff
@@ -162,20 +168,25 @@ class SequentialPipeline : public Pipeline<Input, Output, N> {
...
+        static_assert(std::is_base_of<FunctionStage<typename StageType::InputType,
+                                                    typename StageType::OutputType>, StageType>::value,
+                      "Stage must derive from FunctionStage");
+        StageType *stagePtr = stage.get();
```
nguy8tri: Never do this unless you need the underlying data type. the std::smart_pointer objects override the operator `->` for you so you don't need this.

src/common/pipeline/pipelines.hpp
```diff
@@ -195,10 +206,13 @@ class SequentialPipeline : public Pipeline<Input, Output, N> {
      * @pre This method is called when the number of registered stages is less
      * than N
      */
-    template<typename I> SequentialPipeline &Complete(FunctionStage<I, Output> &stage) {
+    template<typename StageType>
```
nguy8tri: Same comments here.

src/common/pipeline/stages.hpp
```diff
@@ -55,6 +56,15 @@ class Stage : public Action {
 template<typename Input, typename Output>
 class FunctionStage : public Stage<const raw_type<Input> &, raw_type<Output>> {
  public:
+   /**
+    * Canonical, non-reference input type for this stage.
+    */
+    using InputType = raw_type<Input>;
```
nguy8tri: why do we need these? If it was for the type definitions before, I don't think you need these anymore.

src/distance/edge-filters.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <optional>
+#include <utility>
+
+#include "command-line/parsing/options.hpp"
+#include "common/pipeline/pipelines.hpp"
+#include "distance/edge.hpp"
+
+namespace found {
+
+
```
nguy8tri: Get rid of 2 new lines

src/providers/factory.hpp
```diff
@@ -29,10 +31,23 @@ inline std::unique_ptr<CalibrationPipelineExecutor> CreateCalibrationPipelineExe
...
+    std::unique_ptr<DistanceDeterminationAlgorithm> distAlg = ProvideDistanceDeterminationAlgorithm(std::move(options));
+    std::unique_ptr<VectorGenerationAlgorithm> vecAlg = ProvideVectorGenerationAlgorithm(std::move(options));
+
+    if (filtersOpt) {
+        return std::make_unique<DistancePipelineExecutor>(std::move(options),
+                                    std::move(edgeAlg),
+                                    std::move(filtersOpt),
+                                    std::move(distAlg),
+                                    std::move(vecAlg));
+    } else {
```
nguy8tri: Get rid of else clause

src/providers/factory.hpp
```diff
@@ -29,10 +31,23 @@ inline std::unique_ptr<CalibrationPipelineExecutor> CreateCalibrationPipelineExe
  * @return A pointer to a DistancePipelineExecutor
  */
 inline std::unique_ptr<DistancePipelineExecutor> CreateDistancePipelineExecutor(DistanceOptions &&options) {
-    return std::make_unique<DistancePipelineExecutor>(std::forward<DistanceOptions>(options),
-                                    ProvideEdgeDetectionAlgorithm(std::forward<DistanceOptions>(options)),
-                                    ProvideDistanceDeterminationAlgorithm(std::forward<DistanceOptions>(options)),
-                                    ProvideVectorGenerationAlgorithm(std::forward<DistanceOptions>(options)));
+    std::unique_ptr<EdgeDetectionAlgorithm> edgeAlg = ProvideEdgeDetectionAlgorithm(std::move(options));
```
nguy8tri: Use perfect forwarding here via `std::forward`

src/providers/factory.hpp
```diff
@@ -29,10 +31,23 @@ inline std::unique_ptr<CalibrationPipelineExecutor> CreateCalibrationPipelineExe
...
+    std::unique_ptr<EdgeFilteringAlgorithms> filtersOpt = ProvideEdgeFilteringAlgorithm(std::move(options));
+    std::unique_ptr<DistanceDeterminationAlgorithm> distAlg = ProvideDistanceDeterminationAlgorithm(std::move(options));
+    std::unique_ptr<VectorGenerationAlgorithm> vecAlg = ProvideVectorGenerationAlgorithm(std::move(options));
+
+    if (filtersOpt) {
+        return std::make_unique<DistancePipelineExecutor>(std::move(options),
```
nguy8tri, line 44: Forward options

src/providers/factory.hpp
```diff
@@ -29,10 +31,23 @@ inline std::unique_ptr<CalibrationPipelineExecutor> CreateCalibrationPipelineExe
...
+                                    std::move(edgeAlg),
+                                    std::move(filtersOpt),
+                                    std::move(distAlg),
+                                    std::move(vecAlg));
+    } else {
+        return std::make_unique<DistancePipelineExecutor>(std::move(options),
```
nguy8tri: Forward options

src/common/pipeline/pipelines.hpp
nguy8tri: If you want, you can change AddStage and Complete to accept && parameters to save space, but I'll let you decide.

test/common/mocks/distance-mocks.hpp
```diff
@@ -24,6 +24,13 @@ class MockVectorGenerationAlgorithm : public VectorGenerationAlgorithm {
     MOCK_METHOD(PositionVector, Run, (const PositionVector& points), (override));
 };
 
+class MockEdgeFilteringAlgorithm : public EdgeFilteringAlgorithm {
```
nguy8tri: Place in order (so below edge detection)

test/distance/edge-filters-test.cpp
```diff
@@ -0,0 +1,44 @@
+#include <gtest/gtest.h>
```
nguy8tri, line 1: Not sure what this is here for, perhaps you should put tests in here for NoOpEdgeFilter instead, because you're testing pipeline behavior as opposed to actual filter behavior. In other words, a test for using the no op edge filtering algorithm will suffice here.

test/integration/integration-test.cpp
```diff
@@ -309,5 +309,57 @@ TEST_F(IntegrationTest, TestCalibrationDistanceCombinedPipelineOtherOutput) {
...
+    ASSERT_GE(DEFAULT_MAG_ERR_TOL,
+              (example_earth1.position.Magnitude() - actual.positions[0].position.Magnitude())
+                / example_earth1.position.Magnitude());
+    ASSERT_GE(DEFAULT_ARC_SEC_TOL, RadToArcSec(Angle(example_earth1.position, actual.positions[0].position)));
+
+    std::remove(temp_df);
+}
+
+// TODO: Remove/replace when noop-edge-filter is not needed.
+TEST_F(IntegrationTest, TestMainDistanceNoOpEdgeFilterDisabled) {
```
nguy8tri: Doesn't one of the other tests cover not using the filter? In that case you could get rid of this.

nguy8tri (changes requested):
Good job on the edits. I have now started moving towards stylistic edits (with a few programmatic ones too).

src/common/pipeline/pipelines.hpp
MahirEmran: I think I'm a bit confused here, how would this save space? Theyr'e references so we aren't copying anyways. Also if we did && then the caller could just give a temporary stage (which then could mean the stage might not outlive pipeline)

nguy8tri (approved):
Everything looks good, congratulations on doing this

PR #49, opened 2025-10-31
Optimization/Enhancement: Iterative Spherical Distance Determination Algorithm (ISDDA) and Spherical Distance Determination Algorithm (SDDA)

### LINKED ISSUE #27: Enhancement/Optimization: Iterative Spherical Distance Determination Algorithm (ISDDA)
# Description

1. Optimize ISDDA to no longer rely on heap memory. The variables `losses` and `positions` are used to calculate an iterative, weighted sum. We only really need to keep track of the first loss and first position, and create variables that accumulate loss and weighted positions, i.e. we want to make `decimal totalLoss` and `Vec3 totalPosition` objects that accumulate in the following manner in the loop:
- `loss_i = this->GenerateLoss(position, targetDistSq, targetRSq, projectedPoints, pointsSize) / losses[0]`
- `PositionVector position_i(SphericalDistanceDeterminationAlgorithm::Run({p[indicies[j]], p[indicies[j + 1]], p[indicies[j + 2]]}));`
- `totalLoss += DECIMAL_EXP(-loss_i)`
- `totalPosition += position_i * DECIMAL_EXP(-loss_i)`

Allowing us to get rid of `losses` and `positions` and their heap-allocated data. We then return:

`totalPosition / totalLoss`

> [!WARNING]
> We only consider a loss/position pair when loss is less than the discriminator ratio. We otherwise throw them out. We find this in the line:
> ```C++ 
> if (losses[i] <= this->discriminatorRatio_) positions[i++] = position;
> ```
> You will still need to keep track of `i` of course.

2. Allow for the refresh of the reference position and loss (what currently is the very first position and loss). Add a parameter to the constructor of ISDDA that indicates how many times to refresh the reference position. Then, you should calculate at what iteration numbers you should refresh the position to fulfill this. An easy way to do this is to do `uint64_t refresh_iteration = num_iterations_/refresh_rate_` and the condition of refresh is when `i % refresh_iteration == 0` (this does refresh on `i == 0`, but it will have no effect then if `totalLoss` and `totalPosition` are configured right. To refresh, you simply perform `ref_position = totalPosition / totalLoss` and generate the loss on it to get the reference loss.
    - I theorize this will boost the accuracy of the algorithm and cause it to converge earlier, but you will need to confirm this.
    - To be honest, an alternate strategy is to, after establishing the initial position and loss, create another loop around the while loop and return statement that runs for the number of refresh times, and instead of returning, set the initial position to the `totalPosition / totalLoss` and recalculate the initial loss, and that will be eaiser. Thus the outer loop runs for the number of "epochs" and the inner loop runs for the number of "iterations per epoch". If you chose this, you should name those variables to these descriptions, and its more consistent with usual ML anyways.

nguy8tri:
# Description
This PR optimizes and updates ISDDA to increase performance, in addition to modifying and optimizing SDDA. This PR closes #27.

## Changes
1. Optimized SDDA performance by getting rid of unnecessary SQRT usage
2. Added fields to SDDA to make intermediate calculations available to ISDDA
3. Optimized memory usage by:
    - removing unnecessary heap-allocated data structures,
    - Optimized memory copies via optimizing argument passing to SDDA's Run function.
4. Added refresh option (maximumRefreshes) to reference position in ISDDA (to support moving target)
5. Added clearer comments on ISDDA
6. Optimized ISDDA runtime by eliminating shuffle indices.

### Minor Changes
1. Removed `PreciseDecimal`
2. Made test cases more tolerable to differences in ISDDA calculations
3. Changed workflow so Documentation no longer depends on other targets, and allows coverage to pass so we can examine the coverage via the artifact link
4. Changed workflow so to use simple clean (to save time)

## Testing
1. Updated tests to reflect new maximumRefreshes parameter and command-line argument for ISDDA

# Artifacts for PR #49 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/49/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/49/merge/doxygen)


test/integration/integration-test.cpp
```diff
@@ -20,9 +20,9 @@
 namespace found {
 
 /// The default arc second tolerance
-#define DEFAULT_ARC_SEC_TOL DECIMAL(1500)  // Equivalent to 5/12 of a degree
+#define DEFAULT_ARC_SEC_TOL DECIMAL(2000)  // Equivalent to 5/9 of a degree
 /// The default magnitude error tolerance
-#define DEFAULT_MAG_ERR_TOL DECIMAL(0.01)  // Equivalent to 1%
+#define DEFAULT_MAG_ERR_TOL DECIMAL(0.02)  // Equivalent to 2%
```
j4lando: I am surprised that getting rid of precise decimal makes it this much more inaccurate. Also do we want to make a change that makes our algorithm worse.

j4lando (commented):
I am unsure if getting rid of precise decimal entirely is the correct choice if it decreases are accuracy by a whole percentage point which equates to a couple of km

test/integration/integration-test.cpp
nguy8tri: Its not that, its that I introduced another hyperparameter into the algorithm (ISDDA: maxRefreshes), which destablized the variance on its results. I'll change it back though because I thought I had to change it initially to make it more permissive in accounting for that hyperparamater, however, I had already changed it to a more moderate value (one)

nguy8tri: So, I'll revert the accuracy change and you'll observe it still passes.

j4lando (approved):
Looks good.

PR #60, opened 2025-11-15
Optimization with Time Flag

### LINKED ISSUE #44: Enhancement/Optimization: Time Flag
FOUND is not necessarily run at the same time an image is taken. Therefore time should be passed in as a flag so that the algorithm uses the time at which the picture was taken instead of the time at runtime.

Eesha-Jain:
# Description
Resolves https://github.com/UWCubeSat/found/issues/44

Adds a time flag request to found's distance command that marks when the image was taken (in nanoseconds). This allows for a more accurate understanding of the satellite's latitude and longitude. Currently outputs time into the outputted .found file.

This PR also modifies the output of output.cpp's GetEarthCoordinates function to be ECEF coordinates. This is the standard set for representing position by NASA (example: [https://ntrs.nasa.gov/api/citations/20190002515/downloads/20190002515.pdf](https://ntrs.nasa.gov/api/citations/20190002515/downloads/20190002515.pdf)) and other space grade receivers. 

Note that the nanoseconds since epoch is being stored as an unsigned 64-bit integer throughout the code. This can reliably store up until the year 2557.

# Artifacts for PR #60 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/60/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/60/merge/doxygen)


test/integration/integration-test.cpp
```diff
@@ -123,34 +123,11 @@ TEST_F(IntegrationTest, TestMainCalibrationGeneral) {
     ASSERT_DF_EQ(expected, actual, 1);
 }
 
-TEST_F(IntegrationTest, TestMainDistanceWithManualRelOrientationPrint) {
-    int argc = 6;
```
j4lando, line 127: Why are you deleting this test?

README.md
```diff
@@ -75,18 +75,20 @@ We can then use this calibration information to now make some position estimates
 ```bash
 ./build/bin/found distance                         \
     --image test/common/assets/example_earth1.png  \
+    --image-time "2025-11-11 19:30:00"             \
```
j4lando: Sorry about mentioning this so late in the process, but you're going to want more precision than a second. Our satellite is moving at ~7km/s. I believe time truncates so this means our position could be off by 7km! Since we have a GNSS receiver on board we have access to nanosecond level accuracy for time. 

Ideally a user could pass in a time accurate to a second all the way down to a nanosecond and we would support the conversion of those times to a standard representation.

j4lando: This change has many implications. 
1. Update output.cpp and output.hpp to be compatible with more precise times (where we rotate inertial coordinates into fixed coordinates)
2. Check if our datafile can handle nanoseconds. Currently we're writing time to a 64 bit unsigned integer so hopefully we have enough room for nanoseconds. If we don't you might have to start digging around the encoder. 
3. I'm not sure we should be storing are final coordinates in output.hpp as degrees. Can you do a quick dive into how space grade gnss receivers represent position. That would probably be the standard we want to reproduce. This make our software easily portable to our end users (one-to-one swap with gnss receiver)

Any questions?

src/common/spatial/attitude-utils.hpp
```diff
@@ -228,6 +228,63 @@ class Vec3 {
...
+ * 
+ * ECEF is a Cartesian coordinate system with its origin at Earth's center.
+ * The axes are:
+ * - x-axis: Points to the intersection of the equator and prime meridian (0° latitude, 0° longitude)
+ * - y-axis: Points to 90°E longitude on the equator (0° latitude, 90°E longitude)
+ * - z-axis: Points to the North Pole (90°N latitude)
+ * 
+ * All coordinates are in meters.
+ */
+struct ECEFCoordinates {
```
nguy8tri: Why not just `typedef Vec3 ECEFCoordinates`?

src/common/time/time.cpp
```diff
@@ -41,14 +55,15 @@ DateTime getUT1Time() {
     // provides Delta UT1 values from 2025 to 2026, the average is
     // 0.087497 seconds, so we use that as a rough approximation.
     DateTime now = getUTCTime();
-    now.epochs += AVG_DELTA_UT1;  // Adjust for UT1, this is a rough approximation
-    now.second += AVG_DELTA_UT1;  // Adjust seconds accordingly
+    // Convert AVG_DELTA_UT1 (seconds) to nanoseconds and add
+    now.epochs += static_cast<uint64_t>(AVG_DELTA_UT1 * NS_PER_SEC);
```
nguy8tri: Could also establish the constant to already be in nanoseconds if that's easier

src/common/time/time.hpp
```diff
@@ -10,13 +11,31 @@ namespace found {
 // Average Delta UT1 in seconds, based on USNO data
 #define AVG_DELTA_UT1 0.087497
 
+/// Nanoseconds per second
+constexpr uint64_t NS_PER_SEC = 1000000000ULL;
```
nguy8tri: All constants in this should either be defined as a macro or the constexpr. Use discretion. Also `decimal` should be used here

src/common/decimal.hpp
```diff
@@ -25,21 +25,25 @@
...
+#define DECIMAL_M_1_PI          (DECIMAL(M_1_PI))            /* 1/pi */
+#define DECIMAL_M_2_PI          (DECIMAL(M_2_PI))            /* 2/pi */
+#define DECIMAL_M_2_SQRTPI      (DECIMAL(M_2_SQRTPI))        /* 2/sqrt(pi) */
+#define DECIMAL_M_SQRT2         (DECIMAL(M_SQRT2))           /* sqrt(2) */
+#define DECIMAL_M_SQRT1_2       (DECIMAL(M_SQRT1_2))         /* 1/sqrt(2) */
+#define DECIMAL_M_R_E           (DECIMAL(6371008.7714))      /* Earth's Radius */
+#define DECIMAL_M_A_E           (DECIMAL(6378137.0))         /* WGS84 Semi-major axis (m) */
+#define DECIMAL_M_E2_E          (DECIMAL(0.00669437999014))  /* WGS84 First eccentricity squared */
+#define DECIMAL_M_B_E           (DECIMAL_M_A_E * DECIMAL_SQRT(DECIMAL(1.0) - DECIMAL_M_E2_E))
+        /* WGS84 Semi-minor axis (m) */
```
nguy8tri: Fix spacing so they're all above each other (if possible)

src/distance/output.cpp
```diff
@@ -8,25 +8,19 @@
...
     decimal GMST = std::fmod(DECIMAL_M_PI * gmst / DECIMAL(180.0), 2 * DECIMAL_M_PI);
-    // Figure out Earth's Rotating Frame and express the position in that frame
+
+    // Convert from celestial (inertial) frame to Earth-Centered, Earth-Fixed (ECEF) frame
+    // Standard ECEF convention requires rotation by -GMST around Z-axis.
     Quaternion toEarthRotatingFrame = SphericalToQuaternion(GMST, 0, 0);
     Vec3 position = toEarthRotatingFrame.Rotate(celestialVector);
 
-    // Figure out the right ascension and declination of the vector
-    decimal RA = std::atan2(position.y, position.x);  // Huh, the range is [-PI, PI], not [0, 2PI]. That's convenient
```
nguy8tri: I'd be nice to keep this math somewhere, in case someone requests lat/long instead (that's what the original system was)

src/providers/converters.hpp
```diff
@@ -121,6 +123,69 @@ inline Image strtoimage(const std::string &str) {
...
+    int hour = tm.tm_hour;
+    int minute = tm.tm_min;
+    int second = tm.tm_sec;
+
+    if (second > 59) {
+        throw std::invalid_argument("Invalid second in datetime: " + str);
+    }
+
+    // Validate day of month (considering leap years)
+    int days_in_month[] = {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
```
nguy8tri: constants should be either macros or constexpr

nguy8tri (changes requested):
This is by no means comprehensive, but great job, just left some comments!

src/common/spatial/attitude-utils.hpp
```diff
@@ -228,6 +228,19 @@ class Vec3 {
...
+ * 
+ * ECEF is a Cartesian coordinate system with its origin at Earth's center.
+ * The axes are:
+ * - x-axis: Points to the intersection of the equator and prime meridian (0° latitude, 0° longitude)
+ * - y-axis: Points to 90°E longitude on the equator (0° latitude, 90°E longitude)
+ * - z-axis: Points to the North Pole (90°N latitude)
+ * 
+ * All coordinates are in meters.
+ */
+typedef Vec3 ECEFCoordinates;
```
nguy8tri, line 46: At this point, it probably belongs in style.hpp instead (or possibly the output file, but you should defend either decision you make). Alternately, you could forgo ECEFCoordinates altogether and just use Vec3. The primary reason I use typedefs is to clarify what kind of info is being fed between and as an output from each pipeline, but this is not one of those data types. You could still keep the typedef, but also defend that decision as well.

src/common/time/time.cpp
```diff
@@ -18,17 +20,29 @@
...
+    std::tm* utc_time = std::gmtime(&now_t);
+
+    // Calculate nanoseconds within the current second
+    std::chrono::seconds secs = std::chrono::duration_cast<std::chrono::seconds>(duration);
+    int nanosecond = static_cast<int>((nanos - std::chrono::duration_cast<std::chrono::nanoseconds>(secs)).count());
 
     return {
-        now,
+        epochs_ns,
         static_cast<int>(utc_time->tm_year + 1900),  // tm_year is years since 1900
```
nguy8tri: I'm starting to wonder why we are using `int` for this data type, it probably makes more sense to move to an unsigned data type. However, there could be good reason for it. e.g. if the original data type is also a signed integer, we could keep that. I know you didn't write this code (I did), so I'm pushing this onto you, but since you're modifying this code, I'd like you to consider that.

src/common/time/time.cpp
```diff
@@ -58,35 +73,39 @@ decimal getJulianDateTime(DateTime &time) {
     decimal D = time.day + 1721013.5;
     decimal julianDate = A - B + C + D;
 
-    return julianDate + time.hour / DECIMAL(24.0) +
-           time.minute / DECIMAL(1440.0) + time.second / DECIMAL(86400.0) -
+    return julianDate + time.hour / DECIMAL(HOURS_PER_DAY) +
```
nguy8tri: Thanks for allocating constants for these.

src/common/time/time.cpp
```diff
@@ -58,35 +73,39 @@ decimal getJulianDateTime(DateTime &time) {
...
+           time.minute / DECIMAL(MIN_PER_DAY) + time.second / DECIMAL(SEC_PER_DAY) +
+           time.nanosecond / DECIMAL(NS_PER_DAY) -
            DECIMAL(0.5) * (100 * time.year + time.month > 190002.5 ? 1 : -1) +
```
nguy8tri: The constant 190002.5 is related to all dates after Feburary 28th (the end of the month not counting leap years, and the extra 0.5 is corrective). You can link the source of this formula to https://aa.usno.navy.mil/faq/JD_formula. Anyways, you might also want to make a constant from this (you can deduce this yourself looking at the left hand side of the comparison operator)

src/common/time/time.cpp
```diff
@@ -58,35 +73,39 @@ decimal getJulianDateTime(DateTime &time) {
...
     // Purportedly, the Greenwich Mean Sidereal Time (GMST)
-    // is also 15(18.697374558 + 24.06570982441908 * (JDT - 2451545.0))
+    // is also 15(18.697374558 + 24.06570982441908 * (JDT - J2000_JULIAN_DATE))
     // in degrees
 
     // The below formula is from http://tiny.cc/4wal001
     decimal JDT = getJulianDateTime(time);
-    decimal D_tt = JDT - DECIMAL(2451545.0);  // Julian date - J2000.0
-    decimal t = D_tt / DECIMAL(36525.0);  // Julian centuries since J2000.0
+    decimal D_tt = JDT - DECIMAL(J2000_JULIAN_DATE);  // Julian date - J2000.0
```
nguy8tri: For all macro constants you defined, it will be even more readible if we define the macro constants with the DECIMAL already around them (i.e. `#define J2000_JULIAN_DATE (DECIMAL(...))`)

src/common/time/time.hpp
```diff
@@ -10,13 +11,33 @@ namespace found {
 // Average Delta UT1 in seconds, based on USNO data
 #define AVG_DELTA_UT1 0.087497
 
+/// Nanoseconds per second
+constexpr double NS_PER_SEC = 1000000000ULL;
```
nguy8tri: Because in all cases you use `DECIMAL` with these, use double as the data type. Or better yet, switch to using macros here (saving space at compile time as well)

src/common/decimal.hpp
```diff
@@ -25,21 +25,24 @@
...
+#define DECIMAL_M_PI_4     (DECIMAL(M_PI_4))                                             /* pi/4 */
+#define DECIMAL_M_1_PI     (DECIMAL(M_1_PI))                                             /* 1/pi */
+#define DECIMAL_M_2_PI     (DECIMAL(M_2_PI))                                             /* 2/pi */
+#define DECIMAL_M_2_SQRTPI (DECIMAL(M_2_SQRTPI))                                         /* 2/sqrt(pi) */
+#define DECIMAL_M_SQRT2    (DECIMAL(M_SQRT2))                                            /* sqrt(2) */
+#define DECIMAL_M_SQRT1_2  (DECIMAL(M_SQRT1_2))                                          /* 1/sqrt(2) */
+#define DECIMAL_M_R_E      (DECIMAL(6371008.7714))                                       /* Earth's Radius */
+#define DECIMAL_M_A_E      (DECIMAL(6378137.0))                                          /* WGS84 Semi-major axis (m) */
+#define DECIMAL_M_E2_E     (DECIMAL(0.00669437999014))                                   /* WGS84 eccentricity^2 */
+#define DECIMAL_M_B_E      (DECIMAL_M_A_E * DECIMAL_SQRT(DECIMAL(1.0) - DECIMAL_M_E2_E)) /* WGS84 Semi-minor axis (m) */
```
nguy8tri: Honestly, the fact that we push out a comment like this is less readible. Instead of doing this, revert back to what we had previously (possibly with more spaces) and simply make this line into 2 lines, with the comment at the first line. See if that compiles and is more readible.

src/distance/output.cpp
nguy8tri: This is a moment for code reuse. Compare the `GetEarthCoordinates` with the first 3 lines here.

src/distance/output.hpp
```diff
@@ -6,21 +6,38 @@
...
+    decimal gmst;
+};
+
     /**
-     * Represents an Earth-centric
-     * Spherical Vector with time
-     * information
+     * Converts a celestial vector to Earth-Centered, Earth-Fixed (ECEF) coordinates.
+     * 
+     * @param celestialVector The celestial vector to convert (in meters), must be in the
```
nguy8tri: don't need a `must be ...` here, that's in the preconditions `within @pre`. In general, if you want to hold a consumer to certain conditions for this function, use the `@pre` tag and seperate it from the `@param` description

test/providers/converters-test.cpp
```diff
@@ -151,4 +152,155 @@ TEST(ConvertersTest, TestLocationRecordsDataFile) {
     std::remove(temp_df);
 }
 
+TEST(ConvertersTest, TestDateTimeRecent) {
+    std::string str = "2024-02-29 14:30:45";  // Thu, Feb 29, 2024 (leap year)
+    DateTime dt = strtodatetime(str);
+
+    ASSERT_EQ(2024, dt.year);
```
nguy8tri: Extract into a super ASSET_EQ statement (within common.hpp I believe, or wherever they all are) that you can use a one line statement to asset equals against a date. To reduce the arguments, you may wish to construct an expected DT object, and then assert equals against it directly. You would have to define that equality, either within the DT object (operator==) or via a function that does it for you. I've been sticking to the second to reduce the used code within the code-base (because we don't usually use == within the actual implementation with our custom data types), but a case could be made, especially since the release probably optimizes it away. That said, you'd want to update the other data types to do that, but at that point, you should make another Issue for that.

test/providers/converters-test.cpp
```diff
@@ -151,4 +152,155 @@ TEST(ConvertersTest, TestLocationRecordsDataFile) {
...
+    ASSERT_EQ(2, dt.month);
+    ASSERT_EQ(29, dt.day);
+    ASSERT_EQ(14, dt.hour);
+    ASSERT_EQ(30, dt.minute);
+    ASSERT_EQ(45, dt.second);
+    ASSERT_EQ(0, dt.nanosecond);
+
+    // Calculate expected epoch
+    std::tm tm = {};
```
nguy8tri, line 161: Same here, based on the expected dt object, make an ASSERT macro that will do all this within it.

README.md
```diff
@@ -75,18 +75,20 @@ We can then use this calibration information to now make some position estimates
...
+    --reference-orientation 110,0,0                \
 ```
 
 For the distance flags:
 1. The image is just the path to the image we want to use. In this case, we reference an image we use in our test cases
-2. We then specify our calibration data, which we generated from `tools.generator` (check out [our tools folder](tools) for more information)
-3. We can specify camera parameters, like the focal length (m)
-4. Or the pixel size (m)
-5. We then have our reference orientation, which we need to use in conjunction with our calibration data
+2. The time that the image was taken in UT1 time in the format YYYY-MM-DD HH:MM:SS.NS where hours is in military time
```
nguy8tri: Its already understood this is military time, because of the absence of an AM/PM and because we're talking about UT1, which is usually never dealt with in an AM/PM construct.

nguy8tri (changes requested):
Good job on the fixes. This time, my comments are more towards having you think critically about some code changes you've made. Please respond to my comments as appropriate. Also if it takes a long time for me to review your code, ping me on slack so I can make time for it.

j4lando:
It might be better to replace celestial coordinate with equatorial coordinate since equatorial coordinates are a specific definition of celestial coordinates.

src/common/spatial/attitude-utils.hpp
Eesha-Jain: Since Quaternion is also defined in `attitude-utils.hpp`, wouldn't it be appropriate to define the different coordinate systems in the same file? I do think that it may be worth separating the mathematical definitions like Vec2 and Vec3 from definitions related to the coordinate system, like Attitude, Quaternion, etc. since Vec2 and Vec3 have more broader use cases then just to define the coordinate system.

Additionally, I do think it may be beneficial to keep a typedef ECEFCoordinates because I believe it would be confusing if we are converting between multiple types of coordinate points and all of them are defined with a Vec3.

src/common/time/time.cpp
Eesha-Jain: std::tm defines each of these variables as ints because it utilizes negative numbers to go backward, but we don't particularly need this. I will modify this to utilize an unsigned integer

Eesha-Jain:
It seems that you changed celestial to equatorial in your PR, so I won't implement that here to prevent merge conflicts in the future

src/common/time/time.cpp
```diff
@@ -96,16 +116,19 @@ decimal getGreenwichMeanSiderealTime(DateTime &time) {
...
     #else
         DateTime time = getUT1Time();
         return getGreenwichMeanSiderealTime(time);
     #endif
 }
 
-decimal getGreenwichMeanSiderealTime(std::time_t epochs) {
-    return 15 * (DECIMAL(18.697374558) + DECIMAL(24.06570982441908) * (getJulianDateTime(epochs) - DECIMAL(2451545.0)));
+decimal getGreenwichMeanSiderealTime(uint64_t epochs) {
+    return 15 * (DECIMAL(18.697374558) + DECIMAL(24.06570982441908) *
```
nguy8tri, line 130: We should have constants for these 3

src/providers/converters.hpp
```diff
@@ -22,6 +24,8 @@
 // This is preferable than continuing with invalid data and outputting an unintended
 // result.
 
+constexpr int days_in_month[] = {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
```
nguy8tri: scope to the namespace

src/providers/converters.hpp
```diff
@@ -121,6 +125,69 @@ inline Image strtoimage(const std::string &str) {
...
+    if (std::getline(ss, nanos_str) && nanos_str.size() > 1 && nanos_str[0] == '.') {
+        nanos_str = nanos_str.substr(1);
+        nanos_str.resize(9, '0');  // pad or truncate to 9 digits
+        nanosecond = static_cast<uint64_t>(std::stoul(nanos_str));
+    }
+
+    // Convert to nanoseconds: seconds * NS_PER_SEC + nanoseconds
+    uint64_t epochs_ns = static_cast<uint64_t>(t) * NS_PER_SEC + nanosecond;
+
+    return {
```
nguy8tri, line 152: Are we sure there isn't a library function that will do this for us? Please look into that

src/command-line/parsing/options.hpp
```diff
@@ -46,25 +47,26 @@ FOUND_CLI_OPTION("output-file"          , std::string       , outputFile
...
-FOUND_CLI_OPTION("isdda-radius-loss-order" , int               , ISDDARadLossOrd , 4                          , atoi(optarg)               , kNoDefaultArgument, REQ_ASSIGN, "The Radius Loss Order ISSDA (even int)"                       ) \
-FOUND_CLI_OPTION("output-file"             , std::string       , outputFile      , ""                         , optarg                     , kNoDefaultArgument, REQ_ASSIGN, "The output file (*.found)"                                    ) \
+FOUND_CLI_OPTION("image"                   , found::Image      , image           , {}                          , found::strtoimage(optarg)   , kNoDefaultArgument, REQ_ASSIGN, "The image to process (JPG, PNG, TGA, BMP, PSD, GIF, HDR, PIC)") \
+FOUND_CLI_OPTION("image-time"              , found::DateTime   , imageTime       , {}                          , found::strtodatetime(optarg), kNoDefaultArgument, REQ_ASSIGN, "Image capture time (in UT1 in format YYYY-MM-DD HH:MM:SS.NS)" ) \
```
nguy8tri: Its unclear how many nano seconds, or if its required. It may add value to be flexible about the format, maybe not. Usually libraries will parse a string into a date.

test/command-line/parsing/parser-test.cpp
```diff
@@ -114,6 +115,18 @@ TEST_F(ParserTest, DistanceParserGeneral) {
 
     ASSERT_IMAGE_EQ(expectedImage, options.image);
 
+    // Test Time (epochs in nanoseconds: 1762889400 seconds * NS_PER_SEC)
+    DateTime expectedImageTime{1762889400000000000ULL, 2025, 11, 11, 19, 30, 0};
+    ASSERT_EQ(expectedImageTime.epochs, options.imageTime.epochs);
```
nguy8tri: Make a helper macro that will assert over an entire Date time so we can do something like `ASSET_TIME(expected, actual)`

test/common/common.hpp
```diff
@@ -105,6 +105,16 @@ MATCHER_P(LocationRecordsEqual, expected, "") {
     ASSERT_GE(num, lo) << "Value " << num << " is less than lower bound " << lo; \
     ASSERT_LE(num, hi) << "Value " << num << " is greater than upper bound " << hi;
 
+#define ASSERT_DATETIME_EQ(expected, actual) \
```
nguy8tri, line 134: Ah here it is

README.md
```diff
@@ -75,18 +75,20 @@ We can then use this calibration information to now make some position estimates
 ```bash
 ./build/bin/found distance                         \
     --image test/common/assets/example_earth1.png  \
+    --image-time "2025-11-11 19:30:00.00"          \
     --calibration-data example-df.found            \
     --camera-focal-length 85e-3                    \
     --camera-pixel-size 20e-6                      \
-    --reference-orientation 110,0,0
+    --reference-orientation 110,0,0                \
```
nguy8tri: no, there should not have a line continuation at the last one.

nguy8tri (changes requested):
Good job, just a few more minor comments.

src/command-line/parsing/options.hpp
Eesha-Jain: I can make it so that the format doesn't necessarily require NS (so it can be "YYYY-MM-DD HH:MM:SS" or "YYYY-MM-DD HH:MM:SS.NS"). I think it is still important to include nanoseconds because it could provide us with some precision that's necessary for calculating an accurate position.

Issue #62, opened 2026-02-05
Feature: CCSDS123.0-b-2 Compression


MahirEmran:
This is somewhat based on issue #50. Essentially, we want to be able to downlink images, and I did an analysis a bit ago linked here on the CCSDS123.0-b-2 algorithm: [https://docs.google.com/presentation/d/1LWgH1uPRhg9fvZVEPt0cYsui_d0i07wU8YkXMjUGkhU/edit?slide=id.g3b96997c7d2_0_1174#slide=id.g3b96997c7d2_0_1174](https://docs.google.com/presentation/d/1LWgH1uPRhg9fvZVEPt0cYsui_d0i07wU8YkXMjUGkhU/edit?slide=id.g3b96997c7d2_0_1174#slide=id.g3b96997c7d2_0_1174).

After texting the comms lead, he agreed that it would be best to integrate the compression aspect into FOUND, and then comms handles the downlinking itself. 

I think it is better to still leave Josh's original issue open, since the idea proposed of simulations is interesting. This serves to give an initial baseline compression that just uses the image, which potentially can be expanded upon/improved in the future through other ideas.

For those curious, the algorithm description/manual can be found here: [https://ccsds.org/Pubs/123x0b2e2c3.pdf](https://ccsds.org/Pubs/123x0b2e2c3.pdf)

PR #63, opened 2026-02-05
Feature: CCSDS123.0-B-2 Image Compression

MahirEmran:
# Description
Resolves Issue #62 

Adds a CompressionPipeline with a corresponding executor+provider that has a single FunctionStage. CLI arguments are:
- Path to the image (should be .raw)
- Output directory (ultimately outputs a file called `output.bin` in that directory)
- Absolute Error Limit (AEL), default=0 (basically how much do we want to compress, AEL=0 is lossless, > 0 is partially lossy)
- x,y,z - dimensions of image (z is number of bands)
- dtype - type of integer used (default u8be)

This PR utilizes a modified (modifications were so .raw files could be accepted, which makes our life easier IMO) open-source C implementation of the CCSDS123.0-b-2 algorithm and includes it using CMake. When using hybrid entropy, the binary size would increase significantly (This is due to about ~180kb of encoder tables that are used)
- 702328 -> 1133544 bytes, after basic stripping, 409160 -> 827208 bytes

However, with BA entropy, it's not too bad:
- 702328 -> 805200 bytes, after basic stripping, 409160 -> 474952 bytes. 

Note that the 702328/409160 is from the current version on main branch, and does not account for the additional C++ code written to implement the pipeline.

Note that more stripping would be done in actual release. But I don't think this is too bad (performance overall is pretty comparable between hybrid entropy/BA entropy). We can essentially toggle which kind of entropy we use through changing the tag in CMakeLists.txt without having to change the FOUND repo (yay CMake!). The current toggle is using the BA entropy.

There are some annoying things though here. For example, note when I include the C libraries for the algorithm:
```cpp
extern "C" {
#include "ccsds123_internal.h"  // NOLINT(build/include_subdir)
#include "ccsds123_io.h"  // NOLINT(build/include_subdir)
}
```

I added the flags for linter here because of the way the ccsds123 project was organized, I can't include via subdir without it giving me a bad path. But maybe I just overlooked something silly.

In addition, I believe the annoying issues surrounding the function calls in pipeline tests from Issue #47 are back in code coverage :(. I assumed it was bad practice to use those changes here until those get merged into main.

My final note here is that I'm unsure how useful it is for the CompressionPipeline to have a single stage, rather than just handling the logic in ExecutePipeline (since honestly this is pretty much just a wrapper). I mainly did it for consistency purposes, but it could cause more headache than needed. But, it could also be easier for people to modify in future.

# Artifacts for PR #63 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/63/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/63/merge/doxygen)


j4lando:
I think there would need to be a very strong case to include this pipeline in FOUND instead of the flight software. FOUND's purpose is to cary out optical navigation not data handling. That doesn't mean that this is throw away code. It just means we need to talk about moving it to flight software.

nguy8tri:
I think a case could be made, because this is image compression, which could assist in making image sizes go down for analysis. Maybe not as a standalone pipeline, but integrated as an identity stage before edge detection. You could also think of other cases if you wanted to as well. There is also value of just having it in the flight computer, but it could save computation time having it inside our system as well.

MahirEmran:
I think it would probably fit better in FC software; I don't think the computation time of compressing would save us time on analysis, but maybe I have it backwards. I think overall this compression would be a shared element across LOST/FOUND since the algorithm applies well to both. I don't think it's really useful outside of that though, since this is about image data and not general data. Definitely worth chatting about

PR #64, opened 2026-02-25
Zernike moments

Eesha-Jain:
# Description
Implementing the ZernikeEdgeDetectionAlgorithm, which ngests a pixel level edge detection algorithm to obtain a pixel level approximation of the edge of earth and then performs a sub-pixel edge correction using the moment based edge-orientation and offset. This leverages Zernike polynomials, which is a mathematical pattern defined in polar coordinates. 

This algorithm is based on Christian (2017) "Accurate Planetary Limb Localization" (https://www.researchgate.net/publication/317742756_Accurate_Planetary_Limb_Localization_for_Image-Based_Spacecraft_Navigation)

# Artifacts for PR #64 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/64/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/64/merge/doxygen)


src/distance/edge.hpp
j4lando, line 224: create another edge detection class that takes in points and an image and returns points. Zernike moment will extend this class

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
     // useful fields specific to this algorithm and helper methods
 };
 
+/**
+ * Holds the real and imaginary parts of a complex Zernike moment.
+ */
+struct ZernikeMoment {
```
j4lando: move to style.hpp also rename to complex numbers

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+ public:
+    /**
+     * @brief Constructs a new ZernikeEdgeDetectionAlgorithm
+     * 
+     * @param initialEdgeAlgorithm The algorithm to use for initial edge detection
+     * @param windowSize Size of window around each point (must be odd, typically 5-9)
+     * @param transitionWidth Width of edge transition zone (default: 1.66)
+     */
+    ZernikeEdgeDetectionAlgorithm(
+        std::unique_ptr<EdgeDetectionAlgorithm> initialEdgeAlgorithm,
```
j4lando: remove this stages are connected via the pipeline

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+ * It refines edge positions to sub-pixel accuracy by computing Zernike moments A_11 and A_20
+ * in small windows around initial edge points.
+ *
+ * Based on: Christian (2017) "Accurate Planetary Limb Localization"
+ */
+class ZernikeEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
+ public:
+    /**
+     * @brief Constructs a new ZernikeEdgeDetectionAlgorithm
+     *
```
j4lando: have @pre about the window size being odd

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+     */
+    ZernikeEdgeDetectionAlgorithm(
+        std::unique_ptr<EdgeDetectionAlgorithm> initialEdgeAlgorithm,
+        int windowSize = 7,
+        decimal transitionWidth = DECIMAL(1.66))
+    : initialEdgeAlgorithm_(std::move(initialEdgeAlgorithm)),
+        windowSize_(windowSize),
+        transitionWidth_(transitionWidth) {
+        // Ensure window size is odd
+        if (windowSize_ % 2 == 0) {
```
j4lando: throw an error in run not the constructor. @nguy8tri right?

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+            windowSize_++;
+        }
+    }
+
+    /// @brief Destroys the algorithm
+    virtual ~ZernikeEdgeDetectionAlgorithm() {}
+
+    /// @return Size of window around each point (odd)
+    int windowSize() const { return windowSize_; }
```
j4lando: We do not need to make private constants publicly available with getter method

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+
+    /**
+     * Refines edge positions using Zernike moments sub-pixel detection.
+     *
+     * @param image The image of Earth
+     *
+     * @return The refined edge points with sub-pixel accuracy
+     */
+    Points Run(const Image &image) override;
```
j4lando: remove this

test/distance/edge-test/zernike-performance-test.cpp
Eesha-Jain, line 1: Take the pipeline testing out of this repo. Remove all the changes in CMake. Develop unit tests instead

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+ private:
+    /**
+     * Converts the image to grayscale in double precision for Zernike moment calculations.
+     * This method is only utilized for test images that are in RGB.
+     *
+     * @param image The input image to convert.
+     *
+     * @return A newly allocated array of grayscale intensity values, in row-major order.
+     */
+    static std::unique_ptr<decimal[]> imageToGrayscaleDouble(const Image &image);
```
j4lando: this is not necessary we require image that is passed to be grayscale make this an @pre

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+     * @param centerX X-coordinate of the window center (sub-pixel).
+     * @param centerY Y-coordinate of the window center (sub-pixel).
+     * @param windowSize Side length of the square window (number of pixels).
+     *
+     * @return A newly allocated window of size windowSize * windowSize in row-major order.
+     */
+    static std::unique_ptr<decimal[]> extractWindow(
+        const decimal* imageData,
+        int imageWidth,
+        int imageHeight,
```
j4lando: Only need to take an an Image type as parameter this includes with height and data.

also include @pre that the image is only 1 channel

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+
+    /**
+     * Computes the Zernike polynomial convolution kernels Z_11 and Z_20 for a square unit disk 
+     * mapped to the given window size.
+     *
+     * @param windowSize Side length of the square window (kernel dimensions are windowSize x windowSize).
+     *
+     * @return std::tuple of (kernelZ11Real, kernelZ11Imag, kernelZ20) in row-major order.
+     */
+    static std::tuple<std::unique_ptr<decimal[]>, std::unique_ptr<decimal[]>, std::unique_ptr<decimal[]>>
```
j4lando: make a pair of two complex numbers unless you can find a good name for both of them in which case you should make a struct.

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+     *
+     * @param window Grayscale window data in row-major order (windowSize * windowSize elements).
+     * @param kernelZ11Real Real part of the Z_11 kernel.
+     * @param kernelZ11Imag Imaginary part of the Z_11 kernel.
+     * @param kernelZ20 Z_20 kernel.
+     * @param windowSize Side length of the square window.
+     *
+     * @return std::pair of (A_11, A_20) as ZernikeMoment
+     */
+    static std::pair<ZernikeMoment, ZernikeMoment> computeZernikeMoments(
```
j4lando: preferably struct

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+        const decimal* window,
+        const decimal* kernelZ11Real,
+        const decimal* kernelZ11Imag,
+        const decimal* kernelZ20,
```
j4lando: switch to two complex numbers

src/distance/edge.hpp
```diff
@@ -89,6 +92,180 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+
+    /**
+     * Extracts the edge orientation angle from the complex Zernike moment A_11.
+     * The angle corresponds to the direction perpendicular to the edge in the window.
+     *
+     * @param A11 The complex Zernike moment A_11 (real and imaginary parts).
+     *
+     * @return The edge angle in radians (e.g. atan2(A11.imag, A11.real)).
+     */
+    static decimal extractEdgeAngle(const ZernikeMoment& A11);
```
j4lando: remove all statics above too

test/distance/edge-test/zernike-performance-test.cpp
Eesha-Jain: 1. develop unit tests for each private method 
2. look at simpl-eedge-detection-test.cpp -> aim for 100% code coverage, don't do all teh random edge case images (none, all earth)

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
     // useful fields specific to this algorithm and helper methods
 };
 
+/**
+ * The ZernikeEdgeDetectionAlgorithm class uses Zernike moments for sub-pixel edge detection.
+ * It refines edge positions to sub-pixel accuracy by computing Zernike moments A_11 and A_20
```
j4lando: ZernikeEdgeDetectionALgorithm

1. ingest a pixel level edge detection algorithm to obtain a pixel level approximation of the edge of earth
2. performs a sub-pixel edge correction using the moment based edge-orientation and Offset

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+ * in small windows around initial edge points.
+ *
+ * Based on: Christian (2017) "Accurate Planetary Limb Localization"
+ */
+class ZernikeEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
+ public:
+    /**
+    * Constructs a new ZernikeEdgeDetectionAlgorithm
+    * 
+    * @param initialEdgeAlgorithm_ The algorithm to use for initial edge detection
```
j4lando: initial pixel level edge detection

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+        decimal transitionWidth_ = DECIMAL(1.66))
+        : initialEdgeAlgorithm(std::move(initialEdgeAlgorithm_)),
+        windowSize(windowSize_),
+        transitionWidth(transitionWidth_) {}
+
+    // Destroys the algorithm
+    virtual ~ZernikeEdgeDetectionAlgorithm() {}
+
+    /**
+     * Refines edge positions using Zernike moments sub-pixel detection.
```
j4lando: Run also calls the initial edge algorithm. The description should reflect that run goes from image to pixel level to sub pixel level return

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+     * Solves for the normalized edge distance l in [-1, 1] using the analytical relations
+     * between A'_11, A_20, and l.
+     *
+     * @param A11Prime The rotated Zernike moment A_11 (aligned with the edge direction).
+     * @param A20 The Zernike moment A_20 for the window.
+     * @param transitionWidth The normalized transition width w of the edge model.
+     *
+     * @return The estimated normalized edge distance l
+     */
+    decimal solveEdgeDistance(decimal A11Prime, decimal A20, decimal transitionWidth);
```
j4lando: Do you prefer EdgeDistance or EdgeOffset

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
 }
 
+////// Zernike Edge Detection Algorithm //////
+
+std::unique_ptr<decimal[]> ZernikeEdgeDetectionAlgorithm::extractWindow(
+    const Image &image, const Vec2 &center) {
+    std::unique_ptr<decimal[]> window(new decimal[windowSize * windowSize]);
+    int halfWindow = windowSize / 2;
+
+    // Window is entered around edge pixel
```
j4lando: centered

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+    int centerYInt = static_cast<int>(DECIMAL_ROUND(center.y));
+
+    for (int i = 0; i < windowSize; i++) {
+        for (int j = 0; j < windowSize; j++) {
+            int imgX = centerXInt + (j - halfWindow);
+            int imgY = centerYInt + (i - halfWindow);
+
+            // Handle boundaries: use edge values
+            imgX = std::max(0, std::min(imgX, image.width - 1));
+            imgY = std::max(0, std::min(imgY, image.height - 1));
```
j4lando: I'm a little worried about clamping the edge. Is the Zernike moment calculated from a clamped window valid? 

If not I think it might be best to just discard edges near the boundaries of the image. I believe there are ways to recover these boundary points, but unless they are trivial it might be better to save that for a later pull request.

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+decimal ZernikeEdgeDetectionAlgorithm::extractEdgeAngle(const ComplexNumber& A11) {
+    return DECIMAL_ATAN2(A11.imag, A11.real);
+}
+
+decimal ZernikeEdgeDetectionAlgorithm::solveEdgeDistance(decimal A11Prime, decimal A20, decimal transitionWidth) {
+    if (DECIMAL_ABS(A11Prime) < DECIMAL(1e-10)) {
+        return DECIMAL(0.0);
+    }
+    decimal w = transitionWidth;
+    decimal wSqu = w * w;
```
j4lando: consider storing wSquared instead of w since only wSquared is used

src/distance/edge.cpp
```diff
@@ -366,4 +370,248 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+    decimal A20,
+    decimal transitionWidth) {
+    if (DECIMAL_ABS(A11Prime) < DECIMAL(1e-10)) {
+        return DECIMAL(0.0);
+    }
+    decimal w = transitionWidth;
+    decimal wSqu = w * w;
+
+    decimal discriminant = ((wSqu - DECIMAL(1.0)) * (wSqu - DECIMAL(1.0)) - DECIMAL(2.0) * wSqu * (A20 / A11Prime));
+    return (DECIMAL(1.0) - wSqu - sqrt(discriminant)) / wSqu;
```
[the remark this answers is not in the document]
j4lando: copilot produced valid comment. might want to throw a std::domain_error

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+    }
+
+    Points initialPoints = initialEdgeAlgorithm->Run(image);
+
+    if (initialPoints.empty()) {
+        return initialPoints;
+    }
+
+    // Step 1: Compute Zernike kernels
+    auto [kernelZ11, kernelZ20] = computeZernikeKernels();
```
j4lando: avoid using auto since the type always stays the same. Helps with code readability.

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+            kernelZ11.get(),
+            kernelZ20.get());
+
+        // Step 2c: Extract edge angle ψ from A_11
+        decimal psi = extractEdgeAngle(A11);
+
+        // Step 2d: Rotate A_11 to align with edge direction
+        decimal cosPsi = DECIMAL_COS(psi);
+        decimal sinPsi = DECIMAL_SIN(psi);
+        decimal A11Prime = A11.real * cosPsi + A11.imag * sinPsi;
```
j4lando: Are you sure this is right? This is the equation I found in the book 

Image

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+
+decimal ZernikeEdgeDetectionAlgorithm::extractEdgeAngle(const ComplexNumber& A11) {
+    return DECIMAL_ATAN2(A11.imag, A11.real);
+}
+
+decimal ZernikeEdgeDetectionAlgorithm::solveEdgeDistance(decimal A11Prime, decimal A20, decimal transitionWidth) {
+    if (DECIMAL_ABS(A11Prime) < DECIMAL(1e-10)) {
+        return DECIMAL(0.0);
+    }
+    decimal w = transitionWidth;
```
j4lando: `transitionWidth` is a private variable access it by `this->transitionWidth` instead of passing it as a parameter.

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+     * @return The refined points on Earth's edge in image
+     *
+     * @throws std::invalid_argument if windowSize is not a positive odd integer
+     *
+     * @pre image must only contain one channel
+     */
+    Points Run(const Image &image) override;
+
+ private:
+    /// The initial edge detection algorithm
```
j4lando: add a _ after every variable. This is the convention we use throughout the repo to denote private variables

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+    std::unique_ptr<EdgeDetectionAlgorithm> initialEdgeAlgorithm;
```
j4lando: for example: initialEdgeAlgorithm_

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+        decimal psi = extractEdgeAngle(A11);
+
+        // Step 2d: Rotate A_11 to align with edge direction
+        decimal cosPsi = DECIMAL_COS(psi);
+        decimal sinPsi = DECIMAL_SIN(psi);
+        decimal A11Prime = A11.real * cosPsi + A11.imag * sinPsi;
+
+        // Step 2e: Solve for edge distance l using Christian's equations
+        decimal l = solveEdgeDistance(A11Prime, A20.real, transitionWidth);
+        if (!std::isfinite(static_cast<double>(l))) {
```
j4lando: Consider throwing an error in the solveEdgeDistance. 

You might need a `// GCOVR_EXCL_LINE` since I don't think you should ever end up with a negative discrimant

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+            refinedPoints.push_back(initialPoint);
+            continue;
+        }
+
+        // Step 2f: Convert back to pixel coordinates
+        Vec2 refinedPoint = convertPolarToPixel(initialPoint, l, psi);
+        if (!std::isfinite(static_cast<double>(refinedPoint.x)) ||
```
j4lando: why would this be infinite? Also, don't cast to a `double` alway use `DECIMAL`, but mostly is this logic necessary?

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+        return DECIMAL(0.0);
+    }
+    decimal w = transitionWidth;
+    decimal wSqu = w * w;
+
+    decimal discriminant = ((wSqu - DECIMAL(1.0)) * (wSqu - DECIMAL(1.0)) - DECIMAL(2.0) * wSqu * (A20 / A11Prime));
+    return (DECIMAL(1.0) - wSqu - sqrt(discriminant)) / wSqu;
+}
+
+Vec2 ZernikeEdgeDetectionAlgorithm::convertPolarToPixel(const Vec2& windowCenter, decimal l, decimal psi) {
```
j4lando: consider renaming this function to subpixelEdgeCorrection, applyCorrection, or another name.

It really only converts the correction (l, psi) from polar into pixel coordinates

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+        for (int j = 0; j < windowSize; j++) {
+            kernelZ11[i*windowSize + j].imag =
+                kernelZ11[j*windowSize + i].real;
+        }
+    }
+
+    return {std::move(kernelZ11), std::move(kernelZ20)};
+}
+
+std::pair<ComplexNumber, ComplexNumber> ZernikeEdgeDetectionAlgorithm::computeZernikeMoments(
```
j4lando, line 432: this is called the Frobenius inner product which is a sick name. No change required from this comment. Just wanted to say that.

j4lando:
Nice work making the changes from last time! You're almost there. Hope these comments help. We'll talk soon at the meeting.

src/distance/edge.cpp
Eesha-Jain: :)

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+        ZernikeEdgeDetectionAlgorithm::computeZernikeKernels() {
+    std::unique_ptr<ComplexNumber[]> kernelZ11(new ComplexNumber[windowSize * windowSize]);
+    std::unique_ptr<ComplexNumber[]> kernelZ20(new ComplexNumber[windowSize * windowSize]);
+
+    decimal pixelWidth = DECIMAL(2.0) / DECIMAL(windowSize);
+    decimal offset     = DECIMAL(1.0) - DECIMAL(1.0) / DECIMAL(windowSize);
+
+    constexpr int SUB = 16;
+    decimal subStep   = pixelWidth / DECIMAL(SUB);
+    decimal subOffset = subStep / DECIMAL(2.0);
```
lsakunes: It's not necessary to explicitly cast integers into decimal. Writing `decimal subOffset = subStep / 2;` would work fine. It's cleaner and easier, but it's also (in my opinion) not that big of a deal, so I don't think it's worth the effort to go through and fix every single instance of it.

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
 
+////// Zernike Edge Detection Algorithm //////
+
+std::unique_ptr<decimal[]> ZernikeEdgeDetectionAlgorithm::extractWindow(
+    const Image &image, const Vec2 &center) {
+    std::unique_ptr<decimal[]> window(new decimal[windowSize * windowSize]);
+    int halfWindow = windowSize / 2;
+
+    // Window is entered around edge pixel
+    int centerXInt = static_cast<int>(DECIMAL_ROUND(center.x));
```
nguy8tri: Why the static cast? Is it not possible to use the returned data type?

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
     return result;
 }
 
+////// Zernike Edge Detection Algorithm //////
+
+std::unique_ptr<decimal[]> ZernikeEdgeDetectionAlgorithm::extractWindow(
+    const Image &image, const Vec2 &center) {
```
nguy8tri: Perhaps it might be best to use a `Vec2<int>` might be best, you could parameterize the Vec2 and set the default type parameter `decimal`. You could also consider passing the center as 2 ints.

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+        for (int j = 0; j < windowSize; j++) {
+            int imgX = centerXInt + (j - halfWindow);
+            int imgY = centerYInt + (i - halfWindow);
+
+            // Handle boundaries: use edge values
+            imgX = std::max(0, std::min(imgX, image.width - 1));
+            imgY = std::max(0, std::min(imgY, image.height - 1));
+
+            uint64_t imgIndex = static_cast<uint64_t>(imgY * image.width + imgX);
+            window[i * windowSize + j] = static_cast<decimal>(image.image[imgIndex * image.channels]);
```
nguy8tri: Use `DECIMAL(...)`

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+    for (int i = 0; i < windowSize; i++) {
+        for (int j = 0; j < windowSize; j++) {
+            int imgX = centerXInt + (j - halfWindow);
+            int imgY = centerYInt + (i - halfWindow);
+
+            // Handle boundaries: use edge values
+            imgX = std::max(0, std::min(imgX, image.width - 1));
+            imgY = std::max(0, std::min(imgY, image.height - 1));
+
+            uint64_t imgIndex = static_cast<uint64_t>(imgY * image.width + imgX);
```
nguy8tri: For all integers, use `uint64_t` if you can so we avoid casting.

src/distance/edge.cpp
nguy8tri: The mask is just the pixel?

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+
+std::pair<std::unique_ptr<ComplexNumber[]>, std::unique_ptr<ComplexNumber[]>>
+        ZernikeEdgeDetectionAlgorithm::computeZernikeKernels() {
+    std::unique_ptr<ComplexNumber[]> kernelZ11(new ComplexNumber[windowSize * windowSize]);
+    std::unique_ptr<ComplexNumber[]> kernelZ20(new ComplexNumber[windowSize * windowSize]);
+
+    decimal pixelWidth = DECIMAL(2.0) / DECIMAL(windowSize);
+    decimal offset     = DECIMAL(1.0) - DECIMAL(1.0) / DECIMAL(windowSize);
+
+    constexpr int SUB = 16;
```
nguy8tri: Mind indicating what this 16 means? Also no need to cast it to a decimal. If the first thing in the division operator is a decimal, it will autocast the 16

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+    decimal subStep   = pixelWidth / DECIMAL(SUB);
+    decimal subOffset = subStep / DECIMAL(2.0);
+
```
nguy8tri, line 386: Why are all the numbers up here oddly specific?

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+            decimal intensity = window[idx];
+
+            A11Real += intensity * kernelZ11[idx].real;
+            A11Imag += intensity * kernelZ11[idx].imag;
+            A20Val += intensity * kernelZ20[idx].real;
+        }
+    }
+
+    ComplexNumber A11 = {A11Real, A11Imag};
+    ComplexNumber A20 = {A20Val, DECIMAL(0.0)};  // A_20 is real-only
```
nguy8tri: Can condense the `=` sign, this copies what you've made (So `A11{A11Real, A11Imag}`)

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+                kernelZ11[j*windowSize + i].real;
+        }
+    }
+
+    return {std::move(kernelZ11), std::move(kernelZ20)};
+}
+
+std::pair<ComplexNumber, ComplexNumber> ZernikeEdgeDetectionAlgorithm::computeZernikeMoments(
+    const decimal* window, const ComplexNumber* kernelZ11, const ComplexNumber* kernelZ20) {
+    decimal A11Real = DECIMAL(0.0);
```
nguy8tri: You just need 0

src/distance/edge.cpp
nguy8tri: Again just 0

src/distance/edge.cpp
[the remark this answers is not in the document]
nguy8tri: Try not to throw errors, and handle it gracefully as it's asking you (if you algorithm can tolerate it)

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+    decimal wSqu = w * w;
+
+    decimal discriminant = ((wSqu - DECIMAL(1.0)) * (wSqu - DECIMAL(1.0)) - DECIMAL(2.0) * wSqu * (A20 / A11Prime));
+    return (DECIMAL(1.0) - wSqu - sqrt(discriminant)) / wSqu;
+}
+
+Vec2 ZernikeEdgeDetectionAlgorithm::convertPolarToPixel(const Vec2& windowCenter, decimal l, decimal psi) {
+    decimal scale = DECIMAL(windowSize) / DECIMAL(2.0);
+    decimal offsetX = scale * l * DECIMAL_COS(psi);
+    decimal offsetY = scale * l * DECIMAL_SIN(psi);
```
nguy8tri: why twice? Also could probably be in one line

src/distance/edge.cpp
nguy8tri: Yes, do not use auto

src/distance/edge.cpp
```diff
@@ -366,4 +368,195 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+    Points refinedPoints;
+
+    for (const Vec2& initialPoint : initialPoints) {
+        // Step 2a: Extract window around the point
+        std::unique_ptr<decimal[]> window = extractWindow(image, initialPoint);
+
+        // Step 2b: Compute Zernike moments for this window
+        ComplexNumber A11;
+        ComplexNumber A20;
+        std::tie(A11, A20) = computeZernikeMoments(
```
nguy8tri: If window is only used here, better to just inline `extractWindow`. Also, for your algorithms, use `unique_ptr<X> &` as the parameters, don't ever use `.get()` unless you really need it.

src/distance/edge.cpp
nguy8tri: Shouldn't we chek instead what the value of the transition width is?

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
     // useful fields specific to this algorithm and helper methods
 };
 
+/**
+ * The ZernikeEdgeDetectionAlgorithm class uses Zernike moments for sub-pixel edge detection.
+ * It refines edge positions to sub-pixel accuracy by computing Zernike moments A_11 and A_20
+ * in small windows around initial edge points.
+ *
+ * Based on: Christian (2017) "Accurate Planetary Limb Localization"
```
nguy8tri, line 100: A link might be nice

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+     * Out-of-bounds coordinates are clamped to the image edges.
+     *
+     * @param image The image of the Earth.
+     * @param center The center of the window.
+     *
+     * @return A newly allocated window of size windowSize * windowSize in row-major order.
+     *
+     * @pre image must only contain one channel
+     */
+    std::unique_ptr<decimal[]> extractWindow(const Image &image, const Vec2 &center);
```
nguy8tri: Probably unnecessary

src/distance/edge.hpp
```diff
@@ -89,6 +91,129 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+     * Computes the Zernike moments A_11 and A_20 for a window by convolving the window
+     * with the precomputed Z_11 and Z_20 kernels.
+     *
+     * @param window Grayscale window data in row-major order (windowSize * windowSize elements).
+     * @param kernelZ11 Z_11 kernel at each position in a windowSize * windowSize array
+     * @param kernelZ20 Z_20 kernel at each position in a windowSize * windowSize array
+     *
+     * @return std::pair of (A_11, A_20), each represented as a ComplexNumber
+     */
+    std::pair<ComplexNumber, ComplexNumber> computeZernikeMoments(
```
nguy8tri, line 160: Change to either reference or unique pointer.

nguy8tri (changes requested):
Good job on beginning the implementation of this complicated algorithm. However, your tests are failing, and normally I don't immediately provide feedback, but I think for your case it would be nice. Please fix my workflow before the next thing comes in. Most concerning is the lack of tests - We don't know if your algorithm does what it intends unless we test it.

Mostly some preliminary notes I've created. I'm also more interested in understanding the algorithm. Thanks!

src/distance/edge.cpp
lsakunes: Sorry, I forgot to clarify, it is important to cast to decimal if both of your inputs are integers. For example, line 408 would truncate the result to an integer. However, on line 409, since one of your inputs is a decimal, the cast isn't necessary. I suppose if it gets confusing, it's not the worst thing to cast everything, but I prefer not to.

nguy8tri:
Please also fill out the description

src/distance/edge.cpp
Eesha-Jain: one utilizes COS and the other uses SIN, so I'm not sure if they could be combined into one line

src/distance/edge.cpp
Eesha-Jain: SUB represents the number of rectangles being used in the reinmann sum. since the math is being performed in polar coordinates, the pixel width is being condensed to a width of 2 (from -1 to 1). all the calculations are made accordingly, which is why DECIMAL(2.0) is utilized a lot

src/common/spatial/attitude-utils.cpp
```diff
@@ -12,34 +12,41 @@ namespace found {
 ///////// VECTOR CLASSES //////////
 ///////////////////////////////////
```
nguy8tri, line 14: This will disappear once Eigen is merged, but is necessary for now.

src/common/spatial/attitude-utils.cpp
```diff
@@ -115,7 +122,7 @@ Mat3 Vec3::OuterProduct(const Vec3 &other) const {
 ///// VECTOR UTILITY FUNCTIONS ////
 ///////////////////////////////////
 
-Vec2 Midpoint(const Vec2 &vec1, const Vec2 &vec2) {
+Vec2<> Midpoint(const Vec2<> &vec1, const Vec2<> &vec2) {
```
nguy8tri, line 125: Is it really valid to have empty brackets? Shouldn't there be a template? Is this C++'s version of a wildcard (like in Java)?

src/distance/edge.cpp
```diff
@@ -366,4 +368,177 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+        ZernikeEdgeDetectionAlgorithm::computeZernikeKernels() {
+    const std::size_t n = static_cast<std::size_t>(maskSize_ * maskSize_);
+    std::unique_ptr<ComplexNumber[]> kernelM11 = std::make_unique<ComplexNumber[]>(n);
+    std::unique_ptr<ComplexNumber[]> kernelM20 = std::make_unique<ComplexNumber[]>(n);
+
+    // Normalize window such that the unit circle is inscribed in the window
+    decimal pixelWidth = DECIMAL(2.0) / maskSize_;
+    decimal offset     = DECIMAL(1.0) - DECIMAL(1.0) / maskSize_;
+
+    constexpr int samplesPerPixel = 1028;
```
nguy8tri, line 383: If its in a function I much prefer a macro

src/distance/edge.cpp
```diff
@@ -366,4 +368,177 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+        for (int j = 0; j < maskSize_; j++) {
+            // Center of the pixel in the normalized window
+            decimal uCenter = ((DECIMAL(2.0) * j) / maskSize_) - offset;
+            decimal vCenter = ((DECIMAL(2.0) * i) / maskSize_) - offset;
+
+            // Utilizing midpoint rule to approximate the integral in Eq.(72) with a Riemann sum
+            decimal sum11 = 0;
+            decimal sum20 = 0;
+
+            for (int si = 0; si < samplesPerPixel; si++) {
```
nguy8tri, line 397: Is there a way to do this inner double loop as a simple summation instead of a loop? Is there a way to figure out whether the condition of the pixel being in the unit disk applies? Notice how u and v increase with every iteration. You might be able to figure out what values will work using algebra or something.

src/distance/edge.cpp
```diff
@@ -366,4 +368,177 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
     return result;
 }
 
+////// Zernike Edge Detection Algorithm //////
+
+std::pair<std::unique_ptr<ComplexNumber[]>, std::unique_ptr<ComplexNumber[]>>
+        ZernikeEdgeDetectionAlgorithm::computeZernikeKernels() {
```
nguy8tri, line 374: Have you considered using a set Zernike kernel? Accepting that as an optional parameter might be in your best interest as 1028*1028 is a lot of iterations.

src/distance/edge.cpp
```diff
@@ -366,4 +368,177 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+            A11Real += intensity * kernelM11[idx].real;
+            A11Imag += intensity * kernelM11[idx].imag;
+            A20Val += intensity * kernelM20[idx].real;
+        }
+    }
+
+    return {ComplexNumber{A11Real, A11Imag}, ComplexNumber{A20Val, 0}};  // A_20 is real-only
+}
+
+decimal ZernikeEdgeDetectionAlgorithm::extractEdgeAngle(const ComplexNumber &A11) {
```
nguy8tri, line 457: inline this

src/distance/edge.cpp
```diff
@@ -366,4 +368,177 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+    return DECIMAL_ATAN2(A11.imag, A11.real);  // Eq.(56)
+}
+
+decimal ZernikeEdgeDetectionAlgorithm::extractEdgeOffset(decimal A11Prime, decimal A20) {
```
nguy8tri, line 461: inline this too

src/distance/edge.cpp
```diff
@@ -366,4 +368,177 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
...
+
+    decimal l = (DECIMAL(1.0) - wSqu - DECIMAL_SQRT(discriminant)) / wSqu;  // Eq.(66)
+    if (l < -1) {
+        return -1;
+    }
+
+    return l;
+}
+
+Vec2<int> ZernikeEdgeDetectionAlgorithm::applyEdgeCorrection(const Vec2<int>& maskCenter, decimal l, decimal psi) {
```
nguy8tri, line 481: should be inlined

src/distance/edge.hpp
```diff
@@ -89,6 +92,119 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+     * Extracts a square mask of pixels centered at the given point for Zernike moment computation.
+     *
+     * @param image The image of the Earth.
+     * @param center The center of the mask.
+     * @param kernelM11 M_11 kernel in row-major order (maskSize_ * maskSize_).
+     * @param kernelM20 M_20 kernel in row-major order (maskSize_ * maskSize_).
+     *
+     * @return a pair of ComplexNumbers (A_11, A_20)
+     */
+    std::pair<ComplexNumber, ComplexNumber> computeZernikeMoments(
```
nguy8tri, line 160: This operation is similar, but not the same as a convolution. Might be helpful to add a note that this is the same as "applying" the kernel"

nguy8tri (changes requested):
Good job addressing my comments, I have a few more changes I'd like to request!

src/distance/edge.hpp
```diff
@@ -89,6 +92,119 @@ class LoCEdgeDetectionAlgorithm : public EdgeDetectionAlgorithm {
...
+    * 
+    * @param initialEdgeAlgorithm The algorithm to use for initial pixel level edge detection
+    * @param maskSize Size of the square mask around each point
+    * @param transitionWidth Width of edge transition zone (default: 1.66)
+    *
+    * @pre maskSize must be a positive odd integer.
+    * @pre transitionWidth must be greater than 0
+    */
+    ZernikeEdgeDetectionAlgorithm(
+        EdgeDetectionAlgorithm &initialEdgeAlgorithm,
```
j4lando, line 117: use a && and move to avoid copying the class

j4lando: I take it back

PR #65, opened 2026-02-27
SVD, eigenvalue calculation, general NxM matrix class

lsakunes:
# Description
and some trivial functions for the vector and matrix classes:

row vector*matrix calculation
outer products
inverse (I don't remember if this was here before but I changed it)
cofactor
adjugate
Cramer's rule for solving linear equations

if I have to test all of these before merging I might lose my sanity but I'll do it

# Artifacts for PR #65 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/65/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/65/merge/doxygen)

### DIFF
--- a/src/common/spatial/attitude-utils.cpp
+++ b/src/common/spatial/attitude-utils.cpp
@@ -6,6 +6,12 @@
 
 #include <stdexcept>
 
+#include <vector>
+#include <numeric>
+
+
+#define DEFAULT_TOLERANCE DECIMAL(1e-3)
+
 namespace found {
 
 ///////////////////////////////////
@@ -41,6 +47,13 @@ Vec2 Vec2::operator+(const Vec2 &other) const {
 
 Vec2 Vec2::operator-(const Vec2 &other) const {
     return { x - other.x, y - other.y };
+}
+
+decimal Vec3::At(int i) const {
+    if (i == 0) return x;
+    if (i == 1) return y;
+    if (i == 2) return z;
+    throw std::runtime_error("index out of bounds");
 }
 
 decimal Vec3::Magnitude() const {
@@ -111,6 +124,13 @@ Mat3 Vec3::OuterProduct(const Vec3 &other) const {
     };
 }
 
+Vec3 Vec3::OuterProduct(const Mat3 &other) const {
+    // We implicitly store everything as column vecs,
+    // so we do (v^T*A)^T = A^T*v where v is the column vector
+    // and A is the other matrix
+    return other.Transpose() * (*this);
+}
+
 ///////////////////////////////////
 ///// VECTOR UTILITY FUNCTIONS ////
 ///////////////////////////////////
@@ -135,6 +155,10 @@ decimal Distance(const Vec3 &v1, const Vec3 &v2) {
     return (v1-v2).Magnitude();
 }
 
+decimal SquareDistance(const Vec3 &v1, const Vec3 &v2) {
+    return (v1-v2).MagnitudeSq();
+}
+
 decimal Angle(const Vec3 &vec1, const Vec3 &vec2) {
     return AngleUnit(vec1.Normalize(), vec2.Normalize());
 }
@@ -148,6 +172,43 @@ decimal AngleUnit(const Vec3 &vec1, const Vec3 &vec2) {
 ///////////////////////////////////
 ///////// MATRIX CLASS ////////////
 ///////////////////////////////////
+
+Matrix Matrix::operator*(const Matrix& other) const {
+    if (_cols != other._rows) {
+            throw std::invalid_argument("Matrix dimensions don't match for multiplication");
+        }
+        Matrix result(_rows, other._cols);
+        for (int i = 0; i < _rows; ++i) {
+            for (int j = 0; j < other._cols; ++j) {
+                decimal sum = 0.0;
+                for (int k = 0; k < _cols; ++k) {
+                    sum += Get(i, k) * other.Get(k, j);
+                }
+                result(i, j) = sum;
+            }
+        }
+        return result;
+}
+
+Matrix Matrix::Transpose() const {
+    Matrix result(_cols, _rows);
+        for (int i = 0; i < _rows; ++i) {
+            for (int j = 0; j < _cols; ++j) {
+                result(j, i) = Get(i, j);
+            }
+        }
+        return result;
+}
+
+decimal Matrix::Norm() const {
+    decimal sum = 0.0;
+    for (int i = 0; i < _rows; ++i) {
+            for (int j = 0; j <_cols; ++j) {
+                sum += Get(i,j) * Get(i, j);
+            }
+        }
+        return std::sqrt(sum);
+}
 
 decimal Mat3::At(int i, int j) const {
     return x[3*i+j];
@@ -193,7 +254,7 @@ Vec3 Mat3::operator*(const Vec3 &vec) const {
     return {
         vec.x*At(0,0) + vec.y*At(0,1) + vec.z*At(0,2),
         vec.x*At(1,0) + vec.y*At(1,1) + vec.z*At(1,2),
-        vec.x*At(2,0) + vec.y*At(2,1) + vec.z*At(2,2),
+        vec.x*At(2,0) + vec.y*At(2,1) + vec.z*At(2,2)
     };
 }
 
@@ -205,16 +266,28 @@ Mat3 Mat3::operator*(const decimal &scalar) const {
     };
 }
 
+bool Mat3::operator==(const Mat3 &other) const {
+    for (int i = 0; i < 3; i++) {
+        for (int j = 0; j < 3; j++) {
+            if (abs(other.At(i, j) - At(i, j)) < DEFAULT_TOLERANCE) return false;
+        }
+    }
+    return true;
+}
+
+
 Mat3 Mat3::Transpose() const {
     return {
         At(0,0), At(1,0), At(2,0),
         At(0,1), At(1,1), At(2,1),
-        At(0,2), At(1,2), At(2,2),
+        At(0,2), At(1,2), At(2,2)
     };
 }
 
 Mat3 Mat3::Inverse() const {
     // https://ardoris.wordpress.com/2008/07/18/general-formula-for-the-inverse-of-a-3x3-matrix/
+    decimal determinant = Det();
+    if (determinant == 0) throw std::invalid_argument("Matrix is non-invertible");
     decimal scalar = 1 / Det();
 
     Mat3 res = {
@@ -226,14 +299,46 @@ Mat3 Mat3::Inverse() const {
     return res * scalar;
 }
 
+Mat3 Mat3::Cofactor() const {
+    Mat3 C ={
+        (At(1,1)*At(2,2) - At(1,2)*At(2,1)), (At(1,0)*At(2,2) - At(1,2)*At(2,0)), (At(1,0)*At(2,1) - At(1,1)*At(2,0)),
+        (At(0,1)*At(2,2) - At(0,2)*At(2,1)), (At(0,0)*At(2,2) - At(0,2)*At(2,0)), (At(0,0)*At(2,1) - At(0,1)*At(2,0)),
+        (At(0,1)*At(1,2) - At(0,2)*At(1,1)), (At(0,0)*At(1,2) - At(0,2)*At(1,0)), (At(0,0)*At(1,1) - At(0,1)*At(1,0))
+    };
+
+    return C;
+}
+
+Mat3 Mat3::Adjugate() const {
+    return Cofactor().Transpose();
+}
+
+Vec3 Mat3::SolveCramer(Vec3 b) {
+    Mat3 DX = {
+        b.x, At(0,1), At(0,2),
+        b.y, At(1,1), At(1,2),
+        b.z, At(2,1), At(2,2)
+    };
+    Mat3 DY = {
+        At(0,0), b.x,  At(0,2),
+        At(1,0), b.y,  At(1,2),
+        At(2,0), b.z,  At(2,2)
+    };
+    Mat3 DZ = {
+        At(0,0),   At(0,1), b.x,
+        At(1,0),   At(1,1), b.y,
+        At(2,0),   At(2,1), b.z
+        };
+    Vec3 x = {DX.Det()/Det(), DY.Det()/Det(), DZ.Det()/Det()};
+    return x;
+}
+
+
 /// 3x3 identity matrix
 const Mat3 kIdentityMat3 = {1,0,0,
                             0,1,0,
                             0,0,1};
 
-///////////////////////////////////
-///////// QUATERNION CLASS ////////
-///////////////////////////////////
 
 /**
  * Create a quaternion which represents a rotation of theta around the axis input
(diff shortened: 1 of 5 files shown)


test/common-test/spatial/SVD-test.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    for (int i = 0; i < 3; i++) {
+        for (int j = 0; j < 3; j++) {
+            mat(i,j) = (rando() % modulo)/divisor;
+        }
+    }
+
+    SVDResult result = ComputeSVD(mat, 3);
+
+    Matrix reconstruction = result.U * result.S * result.V.Transpose();
```
j4lando, line 33: This technically only confirms that you have computed a valid matrix decomposition not the SVD. I would prefer if instead of using a random number generator you hard code in values for a 3x3 matrix and some rectangular matrix. You can use this [site](https://www.wolframalpha.com/input/?i=Singular+value+decomposition&a=) to generate-- `expected` .   

However, I think this approach is clever and a reasonable way of tackling the 10x50 case

Also please include an empty matrix edge case as well as any other important edge cases.

src/common/spatial/attitude-utils.hpp
```diff
@@ -137,9 +145,9 @@ class Vec3 {
     // Unit Vector
 
     /**
-     * Provides the magnitude of this Vec2
+     * Returns a vector that is parallel to this with magnitude 1
```
j4lando, line 148: so valid

src/common/spatial/attitude-utils.cpp
```diff
@@ -43,6 +49,13 @@ Vec2 Vec2::operator-(const Vec2 &other) const {
     return { x - other.x, y - other.y };
 }
 
+decimal Vec3::At(int i) const {
+    if (i == 0) return x;
```
nguy8tri, line 53: could be simplified into something like `(decimal * ) &this + i`. A neat trick you can do, but could actually be bad haha.

src/common/spatial/attitude-utils.hpp
```diff
@@ -306,9 +334,96 @@ decimal Angle(const Vec3 &, const Vec3 &);
...
+/**
+ * A Matrix is a mutable object that represents a NxM Matrix
+ * may or may not be copied directly from https://researchdatapod.com/singular-value-decomposition-svd-cpp/ out of laziness
+*/
+class Matrix {
+ private:
+     /// The values in the matrix
+     std::vector<std::vector<decimal>> data;
+     /// The number of rows
+     int _rows;
```
nguy8tri, line 349: customarily we do `rows_` not `_rows`. However, you should be consistent. Either everything ends with a `_` or nothing does (I prefer the latter)

src/common/spatial/attitude-utils.hpp
```diff
@@ -306,9 +334,96 @@ decimal Angle(const Vec3 &, const Vec3 &);
...
 ///////////////////////////////////
 
+/**
+ * A Matrix is a mutable object that represents a NxM Matrix
+ * may or may not be copied directly from https://researchdatapod.com/singular-value-decomposition-svd-cpp/ out of laziness
+*/
+class Matrix {
+ private:
+     /// The values in the matrix
+     std::vector<std::vector<decimal>> data;
```
nguy8tri, line 347: This isn't good. Using a vector here while is nice gets costly over time. Moreover, I see no evidence that the actual dimensions of a Matrix object ever changes. I thus recommend you move to decimal[N][M] where N and M are class parameters (so `template<int N, int M>`, or something like that). The underlying data type actually is better as a 1D array that stores the matrix in row-major order.

src/common/spatial/attitude-utils.hpp
```diff
@@ -306,9 +334,96 @@ decimal Angle(const Vec3 &, const Vec3 &);
...
+     int _cols;
+
+ public:
+     /** 
+      * Constructor initializes matrix with zeros
+      * @param rows the number of rows
+      * @param cols the number of columns
+      */
+     Matrix(int rows, int cols) : _rows(rows), _cols(cols) {
+         data.resize(rows, std::vector<decimal>(cols, 0.0));
```
nguy8tri, line 360: My point above is made. `resize` costs, and can be avoided with brace initialization!

src/common/spatial/attitude-utils.hpp
```diff
@@ -306,9 +334,96 @@ decimal Angle(const Vec3 &, const Vec3 &);
...
+     decimal& operator()(int i, int j) { return data[i][j]; }
+
+     /**
+      * Access the value at i,j using parentheses
+      * @param i the row to access
+      * @param j the column to access
+      * 
+      * @return the value at (i,j) in this matrix
+      */
+     const decimal& operator()(int i, int j) const { return data[i][j]; }
```
nguy8tri, line 388: Why is there two? Also why are we returning a reference to the data? For primitives, references are not necessary and just cost more.

src/common/spatial/attitude-utils.cpp
```diff
@@ -205,16 +266,28 @@ Mat3 Mat3::operator*(const decimal &scalar) const {
...
-        At(0,2), At(1,2), At(2,2),
+        At(0,2), At(1,2), At(2,2)
     };
 }
 
 Mat3 Mat3::Inverse() const {
     // https://ardoris.wordpress.com/2008/07/18/general-formula-for-the-inverse-of-a-3x3-matrix/
+    decimal determinant = Det();
+    if (determinant == 0) throw std::invalid_argument("Matrix is non-invertible");
     decimal scalar = 1 / Det();
```
nguy8tri, line 291: Don't compute it twice

src/common/spatial/attitude-utils.cpp
```diff
@@ -149,6 +173,43 @@ decimal AngleUnit(const Vec3 &vec1, const Vec3 &vec2) {
...
+        for (int i = 0; i < _rows; ++i) {
+            for (int j = 0; j < other._cols; ++j) {
+                decimal sum = 0.0;
+                for (int k = 0; k < _cols; ++k) {
+                    sum += Get(i, k) * other.Get(k, j);
+                }
+                result(i, j) = sum;
+            }
+        }
+        return result;
```
nguy8tri, line 190: I should note this is a very naive and thus slow algorithm. The runtime of this is O(N^3), which is why we usually end up defaulting to matrix libraries which get it to about O(N^2.3). I'd possibly look into that, or simply into optimizing this painful algorithm

src/common/spatial/attitude-utils.cpp
```diff
@@ -149,6 +173,43 @@ decimal AngleUnit(const Vec3 &vec1, const Vec3 &vec2) {
...
+
+Matrix Matrix::Transpose() const {
+    Matrix result(_cols, _rows);
+        for (int i = 0; i < _rows; ++i) {
+            for (int j = 0; j < _cols; ++j) {
+                result(j, i) = Get(i, j);
+            }
+        }
+        return result;
+}
```
nguy8tri, line 201: Idea - Maybe we should just have a transpose flag that tells us how to interpret the elements. That way we don't actually transpose everything.

src/common/spatial/attitude-utils.cpp
```diff
@@ -205,16 +266,28 @@ Mat3 Mat3::operator*(const decimal &scalar) const {
     };
 }
 
+bool Mat3::operator==(const Mat3 &other) const {
```
nguy8tri, line 269: I get the appeal for this, but in the source code itself do we every figure out if a Mat3 is ==? Otherwise, implement this as a matcher within your test cases.

src/common/spatial/attitude-utils.cpp
```diff
@@ -226,14 +299,46 @@ Mat3 Mat3::Inverse() const {
...
 }
 
+Mat3 Mat3::Cofactor() const {
+    Mat3 C ={
+        (At(1,1)*At(2,2) - At(1,2)*At(2,1)), (At(1,0)*At(2,2) - At(1,2)*At(2,0)), (At(1,0)*At(2,1) - At(1,1)*At(2,0)),
+        (At(0,1)*At(2,2) - At(0,2)*At(2,1)), (At(0,0)*At(2,2) - At(0,2)*At(2,0)), (At(0,0)*At(2,1) - At(0,1)*At(2,0)),
+        (At(0,1)*At(1,2) - At(0,2)*At(1,1)), (At(0,0)*At(1,2) - At(0,2)*At(1,0)), (At(0,0)*At(1,1) - At(0,1)*At(1,0))
+    };
+
+    return C;
```
nguy8tri, line 309: inline the return value.

src/common/spatial/attitude-utils.cpp
```diff
@@ -226,14 +299,46 @@ Mat3 Mat3::Inverse() const {
...
+}
+
+Mat3 Mat3::Adjugate() const {
+    return Cofactor().Transpose();
+}
+
+Vec3 Mat3::SolveCramer(Vec3 b) {
```
nguy8tri, line 316: Just `Solve` and note that it uses crammers rule using the `@note` annotation in the `.hpp` file.

src/common/spatial/attitude-utils.cpp
```diff
@@ -226,14 +299,46 @@ Mat3 Mat3::Inverse() const {
...
+
+
 /// 3x3 identity matrix
 const Mat3 kIdentityMat3 = {1,0,0,
                             0,1,0,
                             0,0,1};
 
-///////////////////////////////////
-///////// QUATERNION CLASS ////////
-///////////////////////////////////
```
nguy8tri, line 236: Why did we get rid of this?

src/common/spatial/attitude-utils.cpp
```diff
@@ -377,6 +482,124 @@ Vec3 Attitude::Rotate(const Vec3 &vec) const {
     }
 }
 
+Mat3 Mat3FromCols(Vec3 col1, Vec3 col2, Vec3 col3) {
```
nguy8tri, line 485: Doesn't belong here, this function has nothing to do with rotations.

src/common/spatial/attitude-utils.cpp
```diff
@@ -377,6 +482,124 @@ Vec3 Attitude::Rotate(const Vec3 &vec) const {
...
+void SVDPowerIteration(const Matrix& A,
+                        std::vector<decimal>& v,
+                        decimal& sigma,
+                        std::vector<decimal>& u,
+                        int maxIter,
+                        decimal tol) {
+    int m = A.NumRows();
+    int n = A.NumCols();
+
+    // initialize with 177 in every entry
```
nguy8tri, line 510: Why 177? Why not just 0?

src/common/spatial/attitude-utils.cpp
```diff
@@ -377,6 +482,124 @@ Vec3 Attitude::Rotate(const Vec3 &vec) const {
...
+///////////////////////////////////
+///////// SVD FUNCTIONS ///////////
+///////////////////////////////////
+
+// NOT COPY PASTED from https://researchdatapod.com/singular-value-decomposition-svd-cpp/
+// because i am a responsible coder
+// We're basically just guessing a random V, computing the corresponding U,
+// using that to compute a slightly more accurate V
+// and just repeating it until it works
+void SVDPowerIteration(const Matrix& A,
```
nguy8tri, line 501: From this point, your comments should really go into the `.hpp` file if they are not inside of the implementation.

src/common/spatial/attitude-utils.cpp
```diff
@@ -377,6 +482,124 @@ Vec3 Attitude::Rotate(const Vec3 &vec) const {
...
+            for (int r = 0; r < m; ++r) {
+                for (int c = 0; c < n; ++c) {
+                    B(r, c) -= sigma * u[r] * v[c];
+                }
+            }
+        }
+        return result;
+}
+
+
```
nguy8tri, line 602: So, whenever you implement an algorithm, the design has to be clear as to why you're choosing the implementation. Because I'm out of the loop, I will ask a few basic questions:
- Is there a reason you chose to do a power iteration to "solve" for SVD? What about other SVD methods? Why not an exact solution?
- How do consumers know how accurate this is? Is it possible to bound the error? (no worries if the answer is no)
- Have you considered using a library for all this?

src/common/spatial/attitude-utils.hpp
nguy8tri: Haha never noticed that.

src/common/spatial/attitude-utils.hpp
nguy8tri: Also, we don't need get AND the operator. One will suffice, or you will simply use the method of one to get the others (there still shouldn't be 2 () operators that do the same thing)

src/common/spatial/attitude-utils.hpp
```diff
@@ -226,6 +234,15 @@ class Vec3 {
...
     Mat3 OuterProduct(const Vec3 &) const;
+
+    /**
+     * Computes row vector * matrix between this and matrix
+     * 
+     * @param other The matrix in this operation
+     * 
+     * @return row vector result of this * other
+    */
+    Vec3 OuterProduct(const Mat3 &) const;
```
nguy8tri, line 245: Not sure what you mean here? Even looking at the implementation doesn't make too much sense what this is really supposed to be. Why is it called "OuterProduct"? I would be more comfortable defining `Mat3::operator*(Vec3)`

src/common/spatial/attitude-utils.hpp
```diff
@@ -318,7 +433,19 @@ class Mat3 {
     /// The matrix entries
     decimal x[9];
 
-    // Accessor
+    /// The symmetric eigenvalue function automatically generates this and it would be a waste to recalculate
+    /// The eigenvalues in descending order
+    Vec3 _eigenvalues;
```
nguy8tri, line 438: What is the point of these? I don't recall even seeing them in the implementation?

src/common/spatial/attitude-utils.hpp
```diff
@@ -700,6 +882,61 @@ class Attitude {
...
+    Matrix U;
+
+    /// Diagonal singular value matrix
+    Matrix S;
+
+    /// Right singular vector matrix
+    Matrix V;
+
+    /**
+     * The constructor
```
nguy8tri, line 904: Not "The constructor`, you should be more descriptive, like `Constructs this SVDResult with a number of rows and columns`. Also, I don't see the point of this constructor. These matricies should be injected into the result, not initialized with a dummy value only to be destroyed soon after. If this is a convenience method for test cases, you should put that initialization into the test cases not here.

src/common/spatial/attitude-utils.hpp
```diff
@@ -700,6 +882,61 @@ class Attitude {
...
 
+
+///////////////////////////////////
+///////// SVD FUNCTIONS ///////////
+///////////////////////////////////
+
+/**
+ * A struct to contain the result of singular value decomposition
+ */
+struct SVDResult {
```
nguy8tri, line 893: Instead, I'd go with `Represents the result of SVD`

src/common/spatial/attitude-utils.hpp
```diff
@@ -722,6 +959,8 @@ class Attitude {
 */
 Mat3 QuaternionToDCM(const Quaternion &);
 
+Mat3 Mat3FromCols(Vec3 col1, Vec3 col2, Vec3 col3);
```
nguy8tri, line 962: Your implementation of this function should go under the above header in the `.cpp` file

test/common-test/spatial/SVD-test.cpp
nguy8tri, line 1: Lowercase the name (so `svd-test.cpp`)

test/common-test/spatial/SVD-test.cpp
nguy8tri: I second this, you will want to generate expected SVD results (even using python) so you can validate this. This also goes for the other tests. Confirming this will in turn confirm that your SVD is a valid reconstruction.

test/datafile/serialization-test.cpp
```diff
@@ -81,7 +81,9 @@ TEST_F(SerializationTest, CorrectHeader) {
 TEST_F(SerializationTest, IncorrectHeader) {
     std::string incorrectStr(reinterpret_cast<const char*>(incorrectCRCTestHeader), sizeof(incorrectCRCTestHeader));
     std::istringstream stream(incorrectStr);
+    testing::internal::CaptureStderr();
```
nguy8tri, line 84: Did we just decide we hated the output of stderr here and in other tests? Please explain why, otherwise, we don't use this functionality unless it is to actually make use of the captured output.

nguy8tri (changes requested):
Thank you Senuka for doing this for the team, it's very much appreciated! I've left some comments everywhere, and I hope you'll be able to make use of them and reflect on the advice I've given you here. This was a good first run of your implementations, now come my wave of nitpicks but also some comments to consider what it is you're actually putting into the code base.

For context, I'd also love to know why we are implementing this algorithm. I'm anticipating something that eventually uses this, but I'm just not sure what that is, and why it necessitates this implementation. Thanks again!

test/datafile/serialization-test.cpp
lsakunes: It causes the github tests to fail

src/common/spatial/attitude-utils.hpp
lsakunes: I forgot about this. The Vec3::operator*(Mat3) does a matrix times a column when it should be a row times a matrix. I implemented the correct version here because I was worried about breaking someone else's code by changing the existing function. I should move this to the original function after checking that no one is using it

src/common/spatial/attitude-utils.hpp
lsakunes: The eigenvectors are used in distance determination, which isn't implemented yet

PR #66, opened 2026-03-03
Eigen

j4lando:
# Description
This pull request sees the rehaul of four major sections of the repository: dev container, linear-algebra, camera, and coordinate definitions. As well as the deletion of the tools section of the repository whose new home is now the found-tools repo. 

## Dev Container
After the python tools were removed from the repository there was no need to include `micro-mamba` in the docker container so it was removed to trim the image size. Update was pushed to docker hub. Additionally the Cmake and C++ extension for visual studio code now auto install upon starting the dev container (prevents reinstalling them every time you reopen the container). 

## Linear Algebra (`attitude-utils.*`)
The decision to switch from the LOST teams linear algebra implementation to the Eigen library was made for several reasons. 

1.  Errors were found in some of the multiplication definitions
2. The Christian Robinson algorithm implements a total least square regression which relies on the SVD which was becoming a burden to implement.
3. Contrary to prior beliefs using the Eigen library decrease binary size and probably speeds up execution time, though a thorough investigation has not been conducted. 

### With Eigen (311K)

image

### Old linear algebra implementation (324K)

image

## Camera
The previous implementation of the camera class did not expose the camera calibration matrix which is useful for many of the calculations for distance determination. 

## Coordinate Definition
The motivation to change the coordinate system originated from the change of camera coordinates but later expanded to a complete overhaul due to the two following oversights of the previous implementation:

1.  declination was defined to be between -180 and 180 degrees instead of -90 to 90 degrees so that sphericalToQuaternion is a one-to-one transformation
2. sphericalToQuaternion now returns a rotation from inputed Euler angles to the global coordinate system. In the context of this project this is a transformation from camera coordinates -> equatorial coordinates. This has the effect of cleaning up a lot of unnecessary conjugations that were taking place throughout the src code. There is now only one .conjugate used in src (`calibrate.cpp`)! However most test now use .conjugate()
In addition to these changes a new flag and variable for the `LOSTVectorGenerationAlgorithm` named `cameraEquatorialCoordinateOffset` was added to account for the discrepancy between the equatorial rotation specified by LOST and the definition of the camera coordinate system. 

## vectorize-test
many of the old vectorize test where removed because they actually tested the quaternion implementation in `attitude-utils.cpp` instead of `vectorize.*`. 

Some small changes that were made:

1.  switched celestial to equatorial (equatorial is the more specific definition of a celestial coordinate system that we use)

# Artifacts for PR #66 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/66/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/66/merge/doxygen)


j4lando:
For some reason I can't edit my PR description. 

At the end I increased the tolerances for float_mode so the test wouldn't fail. A little disturbing I had to increase the tolerance for distance test from 600 to 1400. I thought this was due to the instability caused by the gimbal lock at the poles for `quaternoinToSpherical`, but that function isn't used in the distance test. I'm lowkey stumped and it's probably very bad practice to increase the tolerance.

src/common/spatial/camera.cpp
```diff
@@ -11,56 +11,66 @@ namespace found {
...
+      xPixelPitch(xPixelPitch), yPixelPitch(yPixelPitch) {
+    initializeCalibrationMatrixes();
+}
 
-    decimal focalFactor = focalLength/vector.x/pixelSize;
+Camera::Camera(decimal focalLength, decimal pixelSize,
+               int xResolution, int yResolution)
+    : Camera(focalLength, xResolution, yResolution,
+             xResolution / (decimal) 2.0, yResolution / (decimal) 2.0,
+             pixelSize, pixelSize) {}
```
lsakunes, line 29: Shouldn't we initialize the calibration matrices for this constructor too?

src/common/spatial/camera.hpp
```diff
@@ -53,141 +58,90 @@ class Camera {
...
     * 
-    * @note Assumes that X is the depth direction and that it points away 
-    * from the center of the sensor, i.e., any vector (x, 0, 0) will be at 
-    * (xResolution/2, yResolution/2) on the sensor.
+    * @note Assumes that Z is the depth direction and that it points away
+    * from the center of the sensor. Coordinate system: X=right, Y=down, Z=depth.
+    * Any vector (0, 0, z) will be at (xResolution/2, yResolution/2) on the sensor.
     */
-    Vec2 SpatialToCamera(const Vec3 &) const;
+    Vec2 SpatialToPixelCoordinates(const Vec3 &) const;
```
lsakunes: I like this new convention, I was pretty confused by the old name

test/common/common.hpp
```diff
@@ -41,33 +45,50 @@ MATCHER_P(LocationRecordsEqual, expected, "") {
 
 #define ASSERT_DECIMAL_EQ_DEFAULT(val1, val2) ASSERT_DECIMAL_EQ(val1, val2, DEFAULT_TOLERANCE)
 
+#define ASSERT_ANGLE_EQ(val1, val2, tolerance) \
+    ASSERT_LT(DECIMAL_M_PI - \
+              DECIMAL_ABS(DECIMAL_M_PI - \
+              DECIMAL_ABS(std::fmod(DECIMAL_ABS(val1 - val2), 2*M_PI))) \
+              , tolerance);
```
lsakunes, line 57: fancy

lsakunes (commented):
This is really cool; it looks like it was a lot of work. I am still hesitant to use quaternions but changing everything again would just be wasting time at this point so I think we can switch to Euler only if it becomes a problem. I only had one question about one of the constructors, everything else looked pretty straightforward.

lsakunes:
> For some reason I can't edit my PR description.
> 
> At the end I increased the tolerances for float_mode so the test wouldn't fail. A little disturbing I had to increase the tolerance for distance test from 600 to 1400. I thought this was due to the instability caused by the gimbal lock at the poles for `quaternoinToSpherical`, but that function isn't used in the distance test. I'm lowkey stumped and it's probably very bad practice to increase the tolerance.

@j4lando This vaguely reminds me of the issue I had with CR. Even using a spherical estimation, I don't think we should be getting errors this large to begin with. I'm starting to suspect it's to do with something other than the math, perhaps with the way we handle the numbers. Stil very weird considering doubles should be more precise than this.

src/common/spatial/camera.cpp
lsakunes: I misread the code, I didn't realize it was calling the other constructor. Sorry.

lsakunes (approved):
I think it looks great. If Tri doesn't have any comments, I think this will be very helpful on main.

src/calibrate/calibrate.cpp
```diff
@@ -7,7 +7,7 @@
 namespace found {
 
 Quaternion LOSTCalibrationAlgorithm::Run(const std::pair<EulerAngles, EulerAngles> &orientations) {
-    return SphericalToQuaternion(orientations.first) * SphericalToQuaternion(orientations.second).Conjugate();
+    return SphericalToQuaternion(orientations.first).conjugate() * SphericalToQuaternion(orientations.second);
```
nguy8tri: Oh no, why'd we change this?

src/command-line/parsing/options.hpp
```diff
@@ -46,26 +46,26 @@ FOUND_CLI_OPTION("output-file"          , std::string       , outputFile
...
+FOUND_CLI_OPTION("isdda-radius-loss-order"           , int               , ISDDARadLossOrd                , 4                                                                                         , atoi(optarg)               , kNoDefaultArgument, REQ_ASSIGN, "The Radius Loss Order ISSDA (even int)"                       ) \
+FOUND_CLI_OPTION("output-file"                       , std::string       , outputFile                     , ""                                                                                        , optarg                     , kNoDefaultArgument, REQ_ASSIGN, "The output file (*.found)"                                    ) \
+FOUND_CLI_OPTION("camera-equatorial-coordinate-offset", found::Quaternion , cameraEquatorialCoordinateOffset, found::Quaternion(found::AngleAxis(DECIMAL_M_PI + DECIMAL_M_PI_2, found::Vec3(0, 0, 1))), found::strtoquat(optarg)   , kNoDefaultArgument, REQ_ASSIGN, "The rotation from the equatorial frame into the camera coordinate definition")
```
nguy8tri: This doesn't seem to be the same spacing

src/common/spatial/camera.hpp
```diff
@@ -53,141 +58,90 @@ class Camera {
...
-    * @return A vector in 3d space corresponding to the given vector, with x-component equal to 1
+    * @return A vector in 3d space corresponding to the given vector, with z-component equal to 1
     * 
     * @note Not all vectors returned by this function will necessarily have the same magnitude.
     * 
-    * @warning Other functions rely on the fact that returned vectors are placed one unit away (x-component equal to 1). Don't change this behavior!
+    * @warning Other functions rely on the fact that returned vectors are placed one unit away 
+    * (z-component equal to 1). Don't change this behavior!
     */
-    Vec3 CameraToSpatial(const Vec2 &) const;
```
nguy8tri, line 88: Notice how the names were the opposite of each other? We want that

src/common/spatial/camera.hpp
```diff
@@ -53,141 +58,90 @@ class Camera {
...
     /**
-    * Gives a point in 3d space that could correspond to the given vector, using the same
-    * coordinate system described for SpatialToCamera.
+    * Maps a 2D point in pixel coordinates (row and column index of image matrix) to
+    * a point on the Image plane located at z=1 in 3D space.
     * 
     * @param vector The vector on the camera to convert to a 3D vector
     * 
-    * @return A vector in 3d space corresponding to the given vector, with x-component equal to 1
+    * @return A vector in 3d space corresponding to the given vector, with z-component equal to 1
```
nguy8tri: Why change to Z?

src/common/spatial/camera.cpp
```diff
@@ -11,56 +11,66 @@ namespace found {
...
-    // TODO: is there any sort of accuracy problem when vector.y and vector.z are small?
+Camera::Camera(decimal focalLength,
+               int xResolution, int yResolution,
+               decimal xCenter, decimal yCenter,
+               decimal xPixelPitch, decimal yPixelPitch)
+    : focalLength(focalLength),
+      xResolution(xResolution), yResolution(yResolution),
+      xCenter(xCenter), yCenter(yCenter),
+      xPixelPitch(xPixelPitch), yPixelPitch(yPixelPitch) {
+    initializeCalibrationMatrixes();
```
nguy8tri: Instead of this, you should be able to add, to the end of the list, `calibrationMatrix(initializeCalibrationMatrix(...)), inverse(initializeInverse(...))`

src/common/spatial/camera.cpp
```diff
@@ -11,56 +11,66 @@ namespace found {
...
+}
 
-    decimal focalFactor = focalLength/vector.x/pixelSize;
+Camera::Camera(decimal focalLength, decimal pixelSize,
+               int xResolution, int yResolution)
+    : Camera(focalLength, xResolution, yResolution,
+             xResolution / (decimal) 2.0, yResolution / (decimal) 2.0,
```
nguy8tri: do we need decimal casting here? If we do, use `DECIMAL`

src/distance/distance.cpp
```diff
@@ -17,11 +17,11 @@ namespace found {
 ///// SphericalDistanceDeterminationAlgorithm /////
 
 PositionVector SphericalDistanceDeterminationAlgorithm::Run(const Points &p) {
-    if (p.size() < 3) return {0, 0, 0};
+    if (p.size() < 3) return Vec3(0, 0, 0);
```
nguy8tri, line 20: brace initialization is great, I'd like to know why you changed here?

src/providers/converters.hpp
```diff
@@ -160,7 +191,7 @@ inline LocationRecords strtolr(const std::string &str) {
     while (std::getline(file, line)) {
         std::istringstream iss(line);
         LocationRecord record;
-        if (!(iss >> record.timestamp >> record.position.x >> record.position.y >> record.position.z)) {
+        if (!(iss >> record.timestamp >> record.position.x() >> record.position.y() >> record.position.z())) {
```
nguy8tri, line 163: wait that works? does this return references to x y and z? Otherwise, is the vector even being set?

test/calibrate/calibrate-test.cpp
```diff
@@ -12,60 +12,75 @@
...
+    EulerAngles local = QuaternionToSpherical(Quaternion(0.36, 0.48, 0.64, 0.48));
     EulerAngles reference(0, 0, 0);
 
     LOSTCalibrationAlgorithm algorithm;
-    EulerAngles actual = algorithm.Run(std::make_pair(local, reference)).Canonicalize().ToSpherical();
+    Quaternion result = algorithm.Run(std::make_pair(local, reference));
 
-    ASSERT_EA_EQ_DEFAULT(local, actual);
+    // Round-trip: S2Q(reference) * result.conjugate() should recover S2Q(local)
+    Quaternion roundTrip = SphericalToQuaternion(reference) * result.conjugate();
```
nguy8tri: why has the order flipped across all invocations?

test/calibrate/calibrate-test.cpp
```diff
@@ -12,60 +12,75 @@
...
+    // Compute what the local world→camera quaternion would be for a different reference
+    Quaternion diffLocal = diffReference * result.conjugate();
+
+    // The calibration quaternion should be the same regardless of reference
+    Quaternion lhs = SphericalToQuaternion(local).conjugate() * SphericalToQuaternion(reference);
+    Quaternion rhs = diffLocal.conjugate() * diffReference;
+    if (lhs.w() < 0) lhs = Quaternion(-lhs.w(), -lhs.x(), -lhs.y(), -lhs.z());
+    if (rhs.w() < 0) rhs = Quaternion(-rhs.w(), -rhs.x(), -rhs.y(), -rhs.z());
+    ASSERT_QUAT_EQ_DEFAULT(lhs, rhs);
 }
```
nguy8tri, line 66: could condense some of this functionality to the same function. However, the reason I wrote some of my tests is that intuitively, I knew what orientations the quaternion was supposed to provide, so you may want to keep that intuition so we don't loose to what's automatically produced by the results.

test/distance/output-test.cpp
nguy8tri, line 1: why has this been commented out?

test/distance/vectorize-test.cpp
```diff
@@ -45,104 +45,155 @@ TEST(LOSTVectorGenerationAlgorithmTest, TestIdentityTest) {
...
-    Quaternion relativeOrientation = SphericalToQuaternion(DECIMAL_M_PI, 0, 0);
-    LOSTVectorGenerationAlgorithm vectorGen(relativeOrientation, referenceOrientation);
+    Quaternion referenceOrientation = SphericalToQuaternion(0, 0, 0).conjugate();
+    Quaternion relativeOrientation = SphericalToQuaternion(DECIMAL_M_PI, 0, 0).conjugate();
+    LOSTVectorGenerationAlgorithm vectorGen(relativeOrientation, referenceOrientation, Quaternion::Identity());
 
     // Create a PositionVector to test with
     PositionVector x_E = {100.0, 200.0, 300.0};
-    PositionVector actual = vectorGen.Run(-x_E);
+    PositionVector actual = vectorGen.Run(x_E);
```
nguy8tri, line 62: If the original was a conjugate of what should be, why have we done way with negation here?

CMakeLists.txt
```diff
@@ -134,13 +134,30 @@ add_library(${STB_IMAGE} STATIC ${STB_IMAGE_SRC})
...
+    eigen
+    GIT_REPOSITORY https://gitlab.com/libeigen/eigen.git
+    GIT_TAG 3.4.0
+    GIT_SHALLOW ON
+    EXCLUDE_FROM_ALL
+)
+FetchContent_MakeAvailable(eigen)
+
 # Set the source libraries
 set(SRC_LIBS ${STB_IMAGE})
```
nguy8tri: should have a reference to whatever lib eigen forms, presumably ${EIGEN}

nguy8tri (changes requested):
Left some preliminary comments on here, nice job with the changes. I can't say I've caught everything, because it is quite big and I am a bit exhausted right now, but nice job so far.

I mostly question why the behavior of quaternions changed. That's concerning because the tests last time I left it were in the condition they were supposed to be, so we need to make sure they're being done equivalently.

src/common/spatial/attitude-utils.cpp
```diff
@@ -34,59 +34,26 @@ decimal Distance(const Vec3 &v1, const Vec3 &v2) {
 ////// CONVERSION FUNCTIONS ///////
 ///////////////////////////////////
 
-EulerAngles QuaternionToSpherical(const Quaternion &q) {
```
nguy8tri: Reason for function deletion?

src/providers/stage-providers.hpp
```diff
@@ -86,18 +86,21 @@ std::unique_ptr<DistanceDeterminationAlgorithm> ProvideDistanceDeterminationAlgo
...
         return std::make_unique<LOSTVectorGenerationAlgorithm>(options.calibrationData.relative_attitude,
                                                                referenceOrientation);
     } else {
-        Quaternion relativeOrientation = SphericalToQuaternion(options.relOrientation);
         if (options.refAsOrientation) {
             LOG_INFO("Using provided reference orientation for calibration information");
             return std::make_unique<LOSTVectorGenerationAlgorithm>(referenceOrientation);
         }
-        return std::make_unique<LOSTVectorGenerationAlgorithm>(relativeOrientation, referenceOrientation);
+        Quaternion relativeOrientation = SphericalToQuaternion(options.relOrientation).conjugate();
```
j4lando: check this again
```suggestion
        Quaternion relativeOrientation = SphericalToQuaternion(options.relOrientation).conjugate();
```

src/common/spatial/attitude-utils.cpp
```diff
@@ -4,454 +4,56 @@
...
-}
-
-void SpatialToSpherical(const Vec3 &vec, decimal &ra, decimal &de) {
-    ra = atan2(vec.y, vec.x);
-    if (ra < 0)
-        ra += DECIMAL_M_PI*2;
-    de = asin(vec.z);
+    assert(dec >= -DECIMAL_M_PI/2 && dec <= DECIMAL_M_PI/2);
+    // prevent gimbal lock at the polls by only allowing roll
+    assert(DECIMAL_ABS(dec) != DECIMAL_M_PI/2 || ra == 0);
```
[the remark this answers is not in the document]
j4lando: this is actually fine to fail if someone gives us a "nonsencial" rotation

src/common/spatial/camera.hpp
```diff
@@ -53,141 +58,95 @@ class Camera {
...
     * 
     * @return The 2D Vector that represents the 3D vector on the camera
     * 
-    * @note Assumes that X is the depth direction and that it points away 
-    * from the center of the sensor, i.e., any vector (x, 0, 0) will be at 
-    * (xResolution/2, yResolution/2) on the sensor.
+    * @note Any vector (x, 0, 0), x > 0, will be projected onto the principal point of the sensor.
     */
-    Vec2 SpatialToCamera(const Vec3 &) const;
+    Vec2 CameraToPixelCoordinates(const Vec3 &) const;
```
nguy8tri, line 74: Maybe to avoid confusion, we should name this "ProjectOntoCamera" and the other as "ProjectFromCamera"

src/providers/stage-providers.hpp
```diff
@@ -86,18 +86,21 @@ std::unique_ptr<DistanceDeterminationAlgorithm> ProvideDistanceDeterminationAlgo
...
         return std::make_unique<LOSTVectorGenerationAlgorithm>(options.calibrationData.relative_attitude,
                                                                referenceOrientation);
     } else {
-        Quaternion relativeOrientation = SphericalToQuaternion(options.relOrientation);
         if (options.refAsOrientation) {
             LOG_INFO("Using provided reference orientation for calibration information");
             return std::make_unique<LOSTVectorGenerationAlgorithm>(referenceOrientation);
         }
-        return std::make_unique<LOSTVectorGenerationAlgorithm>(relativeOrientation, referenceOrientation);
+        Quaternion relativeOrientation = SphericalToQuaternion(options.relOrientation);
```
nguy8tri: inline variable into constructor, or you can keep for clarity. Either way works.

test/distance/vectorize-test.cpp
```diff
@@ -87,61 +93,133 @@ TEST(LOSTVectorGenerationAlgorithmTest, TestSimpleZRotationTest2) {
     PositionVector x_E = {100.0, 200.0, 300.0};
     PositionVector actual = vectorGen.Run(x_E);
 
-    PositionVector expected = {-x_E.x * DECIMAL_COS(DECIMAL_M_PI / 3) + x_E.y * DECIMAL_SIN(DECIMAL_M_PI / 3),
-                               -x_E.x * DECIMAL_SIN(DECIMAL_M_PI / 3) - x_E.y * DECIMAL_COS(DECIMAL_M_PI / 3),
-                               -x_E.z};
+    // Run(x_E) = -(relOri * refOri * x_E)
+    PositionVector expected = -(relativeOrientation * referenceOrientation * x_E);
```
nguy8tri: No, you can't have a test like this here. You're just running the algorithm, of course its going to match.

test/distance/vectorize-test.cpp
nguy8tri: I see, because the definition of SphericalToQuaternion has changed, this is now the conjugate rotation.

test/distance/vectorize-test.cpp
```diff
@@ -53,10 +59,11 @@ TEST(LOSTVectorGenerationAlgorithmTest, TestIdentityReferenceSimpleTest) {
 
     // Create a PositionVector to test with
     PositionVector x_E = {100.0, 200.0, 300.0};
-    PositionVector actual = vectorGen.Run(-x_E);
+    PositionVector actual = vectorGen.Run(x_E);
 
     // Check if the result is as expected
-    PositionVector expected = {-100, -200.0, 300.0};  // Negate x and y, keep z
+    // Run(x_E) = -(relOri * refOri * x_E)
+    PositionVector expected = -(relativeOrientation * referenceOrientation * x_E);
```
nguy8tri: You can't do this, you must figure out what the expected position actually is. If you want to restore the original test, simply restore line 59, run with the negative x_E, and conjugate the result of `relativeQuaternion` (maybe `referenceQuaternion` too). You can use that strategy throughout the test to match the previous test, because our definition of SphericalToQuaternion now provides a backwards instead of a forwards rotation.

test/calibrate/calibrate-test.cpp
```diff
@@ -12,60 +12,74 @@
...
-    EulerAngles reference(0, 0, 0);
+    EulerAngles local(0, 0, 0);
+    EulerAngles reference(DECIMAL_M_PI / 4, DECIMAL_M_PI / 6, DECIMAL_M_PI / 3);
 
     LOSTCalibrationAlgorithm algorithm;
-    EulerAngles actual = algorithm.Run(std::make_pair(local, reference)).Canonicalize().ToSpherical();
+    Quaternion actual = algorithm.Run(std::make_pair(local, reference));
 
-    ASSERT_EA_EQ_DEFAULT(local, actual);
+    ASSERT_QUAT_EQ_DEFAULT(SphericalToQuaternion(reference).conjugate(), actual);
```
nguy8tri, line 21: To make this test equivalent to the original, swap local and reference, so you don't have to apply a conjugation.

test/calibrate/calibrate-test.cpp
```diff
@@ -12,60 +12,74 @@
...
 }
 
 TEST(CalibrationTest, TestCalibrateRelativeSimple1) {
     EulerAngles local(DECIMAL_M_PI / 4, 0, 0);
     EulerAngles reference(DECIMAL_M_PI / 2, 0, 0);
 
-    // We rotate -PI/4, or 7PI/4
-    EulerAngles expected(7 * DECIMAL_M_PI / 4, 0, 0);
+    // We rotate PI/4
+    EulerAngles expected(DECIMAL_M_PI / 4, 0, 0);
```
nguy8tri, line 29: No, my original statement is correct here. Fundamentally, this should still represent the rotation needed to go from reference to local, which is -PI/4. This will allow you to avoid doing ASSERT_EQ with the conjugate (I think)

test/calibrate/calibrate-test.cpp
```diff
@@ -12,60 +12,74 @@
...
+    EulerAngles local(DECIMAL_M_PI / 4, 0, 0);
+    EulerAngles reference(3 * DECIMAL_M_PI / 4, 0, 0);
 
+    // result: reference <- local
     LOSTCalibrationAlgorithm algorithm;
     Quaternion result = algorithm.Run(std::make_pair(local, reference));
 
-    Quaternion expectedLocalQ = SphericalToQuaternion(local);
-    Quaternion actualLocalQ = result * SphericalToQuaternion(reference);
+    Vec3 localVector(0 , DECIMAL_COS(DECIMAL_M_PI / 4), DECIMAL_SIN(DECIMAL_M_PI / 4));
```
nguy8tri: Note that we can test this in a similar way to above instead of comparing vectors. You should really be doing that instead.

test/calibrate/calibrate-test.cpp
```diff
@@ -12,60 +12,74 @@
...
-    // See if multiplying the reference to the relative orientation
-    // gives the local orientation back.
-    ASSERT_QUAT_EQ_DEFAULT(expectedLocalQ, actualLocalQ);
+TEST(CalibrationTest, TestCalibrateGeneral) {
+    EulerAngles local(DegToRad(120), DegToRad(20), DegToRad(60));
+    EulerAngles reference(DegToRad(330), DegToRad(50), DegToRad(120));
+
+    LOSTCalibrationAlgorithm algorithm;
+    Quaternion result = algorithm.Run(std::make_pair(local, reference));
+    // local -> reference
```
nguy8tri, line 58: wrong statement, this provides reference -> local (L -> F)

test/calibrate/calibrate-test.cpp
```diff
@@ -12,60 +12,74 @@
...
 
-    Quaternion diffReference = SphericalToQuaternion(4.4, 2.9, 0.7);
-    Quaternion diffLocal = DCMToQuaternion(QuaternionToDCM(result) * QuaternionToDCM(diffReference));
+    // world -> reference * reference -> local = world -> local
+    Quaternion actualLocal = SphericalToQuaternion(reference).conjugate() * result.conjugate();
```
nguy8tri: Equivalent to `(result * SphericalToQuaternion(reference)).conjugate(), which is a useful fact for simplifying your assertion below (I think you're smart enough to figure it out)

src/common/spatial/attitude-utils.hpp
```diff
@@ -738,67 +154,32 @@ Quaternion DCMToQuaternion(const Mat3 &);
...
- * @param vec The vector to convert from
- * @param ra The right ascension that will represent the right ascension
- * of vec
- * @param de The declination that will represent the declination of vec
- * 
- * @post This function's output are the parameters ra and de, and those
- * parameters are modified after this function runs.
-*/
-void SpatialToSpherical(const Vec3 &, decimal &ra, decimal &de);
+    { return SphericalToQuaternion(angles.x(), angles.y(), angles.z()); }
```
nguy8tri, line 180: Please add back a `QuaternionToSpherical` function, which was previously inside `Quaternion::ToSpherical`. If there is a built in function in Eigen to do this, take advantage of that.

nguy8tri (changes requested):
Just a few more comments, as much as I'd like to approve at this stage, the test cases need to be reworked to be clearer, and more honest. I also have notes on minor things about the source code.

src/providers/stage-providers.hpp
```diff
@@ -86,18 +86,21 @@ std::unique_ptr<DistanceDeterminationAlgorithm> ProvideDistanceDeterminationAlgo
  * @return std::unique_ptr<VectorGenerationAlgorithm> The vector generation algorithm
  */
 std::unique_ptr<VectorGenerationAlgorithm> ProvideVectorGenerationAlgorithm(DistanceOptions &&options) {
+    // flips the world to camera roation quaternion to a camera to world rotation quaternion,
```
Eesha-Jain: noticed small typo here "roation" -> "rotation"

nguy8tri (approved):
Great job, let's get this in!

PR #68, opened 2026-03-05
TLS

lsakunes:
# Description
Created a simple Total Least Squares implementation using SVD in attitude utils. 

Input: matrix with each row containing [input row vector, expected output]
Output: Vector that dots with each input vector to minimize the error of the outputs

The implementation is pretty simple, it just uses SVD to find a vector closest to the null space

# Artifacts for PR #68 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/68/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/68/merge/doxygen)


nguy8tri (changes requested):
Uh oh, I cannot review this until Josh's code review is merged, because it is both his and your PR. Please have Josh follow up on my comments first.

src/common/spatial/attitude-utils.hpp
```diff
@@ -851,6 +298,29 @@ constexpr decimal ArcSecToRad(decimal arcSec) {
...
+///// LEAST SQUARES FUNCTIONS /////
+///////////////////////////////////
+
+/**
+ * Performs total least squares on the given matrix
+ * using SVD
+ *   
+ * @param data An NxM matrix with each row corresponding to a vector and an expected dot product output
+ * 
+ * i.e rows enumerated by j may contain [position[j].x, position[j].y, position[j].z, 1]
```
j4lando, line 311: these comments specify how you use the TLS in CRA not what an the TLS algorithm does. Try righting more general comments

Issue #70, opened 2026-03-28
Add option for only using Stack memory


MahirEmran:
This is something that needs to be done before FOUND is fit to be run on embedded systems. This is due to the heap inherently being non-deterministic in how much/what memory is available at a given point in time (due to issues like fragmentation with repeated allocations/frees). The solution to this is using the [ETL library](https://www.etlcpp.com/) to replace STL containers and smart pointer implementations to be placed on the stack.
* Smart pointers are complicated here. What ETL does, is essentially make its own mini heap (called a pool) on the stack, that maintains the reference count (for shared ptrs) and ownership (for unique ptrs), where you provide it the number and the type of object you want to have for that pool.
* Another option for smart pointers could be using std::move on the object directly (by value), but I'm not a huge fan of this personally.

However, we should still maintain the ability to use the heap for convenience purposes for personal testing/usage. In addition, we don't (and shouldn't) need to override malloc/free implementations (namely for stb_image, since we just allocate a large buffer at the start of the program and then free it at the end, which is essentially deterministic). We are mainly concerned about allocating/deallocating objects, which is small allocations that happen frequently.

I believe this will mean we need two separate builds (one with heap and one without), and thus more options in build.sh and the github actions. And thus checking coverage on both separately. I think the toggle for ETL container builds should be done through a CMake option.

PR #71, opened 2026-03-28
ETL library for Stack Memory

MahirEmran:
# Description
Resolves Issue #70 (not quite yet still working thru)

Adds a containers.hpp file that essentially creates aliases of containers/pointer classes in the found namespace using either the ETL/STL versions depending on whether the ETL option in CMake is turned on. 
* Code throughout src/ and the tests have been updated to reflect this. There is some coverage that I have excluded/linter lines I've excluded too which I'll have commented on explaining why.

Also adds new build.sh targets (make-etl and cmake-etl). Results of these are stored in build-etl/ and have their own coverage/executables. 

CMakeLists.txt is also modified to add ETL + the building system for build-etl. I had to do some weird stuff with coverage because it was mixing the two. Same w/ Makefile

ETL has actually decreased the FOUND binary size. 400 KB -> 376 KB with ETL after stripping, 400 -> 406 KB without ETL after stripping. Unsure why it has increased without ETL after stripping.

# Artifacts for PR #71 (DO NOT CHANGE)
- [Coverage Artifact](https://uwcubesat.github.io/found/71/merge/coverage)
- [Doxygen Artifact](https://uwcubesat.github.io/found/71/merge/doxygen)


src/providers/factory.hpp
```diff
@@ -28,11 +28,11 @@ inline std::unique_ptr<CalibrationPipelineExecutor> CreateCalibrationPipelineExe
  * 
  * @return A pointer to a DistancePipelineExecutor
  */
-inline std::unique_ptr<DistancePipelineExecutor> CreateDistancePipelineExecutor(DistanceOptions &&options) {
-    return std::make_unique<DistancePipelineExecutor>(std::forward<DistanceOptions>(options),
-                                    ProvideEdgeDetectionAlgorithm(std::forward<DistanceOptions>(options)),
-                                    ProvideDistanceDeterminationAlgorithm(std::forward<DistanceOptions>(options)),
-                                    ProvideVectorGenerationAlgorithm(std::forward<DistanceOptions>(options)));
+inline DistancePipelineExecutorPtr CreateDistancePipelineExecutor(DistanceOptions options) {
```
MahirEmran: Probabyl need to experiment with forwarding a bit more, was having some issues here

build.sh
nguy8tri, line 1: I don't agree with this, people should explicitly pass in the macro to build with ETL

build.sh
```diff
@@ -29,23 +31,45 @@ case "$1" in
...
         ;;
 
+    make-etl)
+        shift
+        CMD="make FOUND_CONTAINER_BACKEND=ETL $*"
+        ;;
+
     clean)
-        CMD="rm -rf build"
+        CMD="rm -rf build build-etl"
```
nguy8tri: If the build is done in ETL, it should write to the build folder, not a separate folder.

src/command-line/execution/executors.cpp
```diff
@@ -1,16 +1,19 @@
...
-#include <cstring>
 
 #include "common/logging.hpp"
 #include "common/time/time.hpp"
 
 namespace found {
 
 CalibrationPipelineExecutor::CalibrationPipelineExecutor(CalibrationOptions &&options,
-                                                         std::unique_ptr<CalibrationAlgorithm> calibrationAlgorithm)
+                                                         unique_ptr<CalibrationAlgorithm, 1>
```
nguy8tri: This repeated one could be avoided if the default integer is 1

src/command-line/execution/executors.cpp
```diff
@@ -41,9 +44,12 @@ DistancePipelineExecutor::~DistancePipelineExecutor() {
 }
 
 DistancePipelineExecutor::DistancePipelineExecutor(DistanceOptions &&options,
-                                                   std::unique_ptr<EdgeDetectionAlgorithm> edgeDetectionAlgorithm,
-                                                   std::unique_ptr<DistanceDeterminationAlgorithm> distanceAlgorithm,
-                                                   std::unique_ptr<VectorGenerationAlgorithm> vectorizationAlgorithm)
+                                                   unique_ptr<EdgeDetectionAlgorithm, 1>
+                                                   edgeDetectionAlgorithm,
```
nguy8tri: I dont like these multiple lines for the type then the parameter

src/command-line/execution/executors.cpp
```diff
@@ -72,13 +78,17 @@ void DistancePipelineExecutor::OutputResults() {
     if (this->options_.calibrationData.header.version != emptyDFVer) {
         outputDF.header = this->options_.calibrationData.header;
         outputDF.relative_attitude = this->options_.calibrationData.relative_attitude;
-        outputDF.positions = std::make_unique<LocationRecord[]>(outputDF.header.num_positions + 1);
-        std::memcpy(outputDF.positions.get(),
-                    this->options_.calibrationData.positions.get(),
-                    outputDF.header.num_positions);
+        if (outputDF.header.num_positions >= FOUND_MAX_LOCATION_RECORDS) {
```
nguy8tri: Are we sure we want to throw an error. Because this is on a satellite, we might benefit from writting the data to a new file and keeping track that they are associated somehow. I'd also like to ask if its even possible to get to this point too, it should be impossible.

src/command-line/execution/executors.cpp
```diff
@@ -1,16 +1,19 @@
...
 
+#include <algorithm>
+#include <cstring>
 #include <memory>
+#include <stdexcept>
 #include <utility>
-#include <cstring>
 
 #include "common/logging.hpp"
 #include "common/time/time.hpp"
```
nguy8tri, line 12: Import what you use, which should involve a reference to the library where the ETL definitions come out of

src/command-line/execution/executors.cpp
```diff
@@ -72,13 +78,17 @@ void DistancePipelineExecutor::OutputResults() {
...
         outputDF.header = this->options_.calibrationData.header;
         outputDF.relative_attitude = this->options_.calibrationData.relative_attitude;
-        outputDF.positions = std::make_unique<LocationRecord[]>(outputDF.header.num_positions + 1);
-        std::memcpy(outputDF.positions.get(),
-                    this->options_.calibrationData.positions.get(),
-                    outputDF.header.num_positions);
+        if (outputDF.header.num_positions >= FOUND_MAX_LOCATION_RECORDS) {
+            throw std::runtime_error("Position record count exceeds FOUND_MAX_LOCATION_RECORDS");
+        }
+        outputDF.positions.resize(outputDF.header.num_positions + 1);
```
nguy8tri, line 99: The use of ETL here might also be excessive, and it might be in our best interest to use the heap here for simplicity.

src/command-line/found-main.cpp
```diff
@@ -21,7 +21,7 @@ int main(int argc, char **argv) {
     }
     std::string command(argv[1]);
 
-    std::unique_ptr<PipelineExecutor> executor;
+    PipelineExecutorPtr executor;
```
nguy8tri: I'd not like to obscure the fact this is a pointer under a name. That's a no-no in general because it makes it difficult for developers to understand what's going on if they're in a hurry

src/common/containers.hpp
```diff
@@ -0,0 +1,77 @@
...
+    template <typename PoolType, typename ObjectType>
+    static pool_deleter Make(PoolType &pool_ref, ObjectType *object_ptr) {
+        return {&pool_ref, object_ptr, [](void *pool_ptr, void *stored_object) {
+                    static_cast<PoolType *>(pool_ptr)->destroy(static_cast<ObjectType *>(stored_object));
+                }};
+    }
+};
+
+template <typename T, size_t N>
+using vector = etl::vector<T, N>;
```
nguy8tri, line 45: this is an alias, and so I'd rather we use `typedef` than `using`, which has broader uses.

src/common/containers.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include "etl/vector.h"
+#else
+#include <cstddef>
+#include <memory>
+#include <utility>
+#include <vector>
+#endif
+
+namespace found {
+
```
nguy8tri, line 42: might be a good idea to make an extra namespace for these so I don't stare at `unique_ptr` and such as being unscoped. Perhaps another 3 letter acronym will do (ask chatgpt, that's how we came up with FOUND).

src/common/etl_config.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef SRC_COMMON_ETL_CONFIG_HPP_
```
nguy8tri: Should just be in the containers file too

src/datafile/serialization.cpp
```diff
@@ -248,6 +249,12 @@ uint32_t calculateCRC32(const void* data, size_t length) {
 }
 
 void serializeDataFile(const DataFile& data, std::ostream& stream) {
```
nguy8tri, line 255: As per my comment earlier, it might be best to make the data file use the heap, since it won't be too concerning (unless you forsee that the files get large). We also could potentially avoid this by dynamically seeing how much entries are in the file and just setting the macro. We should be able to do that, there are multiple ways I can think of too.

src/distance/distance.cpp
```diff
@@ -123,12 +124,14 @@ PositionVector IterativeSphericalDistanceDeterminationAlgorithm::Run(const Point
     size_t i = 0;
     size_t j = 0;
     size_t pointsSize = p.size();
-    std::unique_ptr<Vec3[]> projectedPoints(new Vec3[pointsSize]);
+    vector<Vec3, FOUND_MAX_POINTS> projectedPoints;
+    projectedPoints.resize(pointsSize);
```
nguy8tri, line 128: Okay, I've seen resize many times now. Why the constant resizing? This seems wasteful to me and could probably be avoided. There should be a way to just set this at construction. I also think we should avoid managing the maximum size within our service code wherever we can.

src/distance/distance.hpp
```diff
@@ -226,7 +227,7 @@ class IterativeSphericalDistanceDeterminationAlgorithm : public SphericalDistanc
      */
     decimal GenerateLoss(PositionVector &position,
                          decimal targetDistanceSq,
-                         std::unique_ptr<Vec3[]> &projectedPoints,
+                         vector<Vec3, FOUND_MAX_POINTS> &projectedPoints,
```
nguy8tri: On second thought, when the mode is set to dynamic, this could be wasteful, as this is forcing vectors to support whatever FOUND_MAX_POINTS is instead of the default, which is much smaller.

src/distance/edge.cpp
```diff
@@ -19,16 +21,22 @@ namespace found {
...
     // Step 0: Define Common Variables
     uint64_t imageSize = image.width * image.height;
+    if (imageSize > FOUND_MAX_IMAGE_PIXELS) {
+        throw std::runtime_error("Image pixel count exceeds FOUND_MAX_IMAGE_PIXELS");
+    }
 
     // Step 1: Obtain the component that represents space
-    Components spaces = ConnectedComponentsAlgorithm(image, [&](uint64_t index, const Image &image) {
-        // Average the pixel, then threshold it
+    std::function<bool(uint64_t, const Image &)> isSpacePixel = [&](uint64_t index, const Image &inputImage) {
```
nguy8tri, line 29: Avoid the construction of a function and saving it if possible. Its actually standard to define lambdas inline rather than as functions.

src/distance/edge.cpp
```diff
@@ -98,6 +106,9 @@ Points SimpleEdgeDetectionAlgorithm::Run(const Image &image) {
             while (points.find(index) == points.end()) index += update;
             if (index / image.width != edge_condition) {
                 index -= update;
+                if (result.size() >= FOUND_MAX_POINTS) {
+                    throw std::runtime_error("Edge point count exceeds FOUND_MAX_POINTS");
```
nguy8tri: Again, this is concerning if we must throw an error. Either the user should account for this at compile time, or we should truncate. In the case of edge detection, we don't need all the points, but this may potentially bias the results. Please weigh the possible options.

src/distance/edge.cpp
```diff
@@ -279,50 +298,60 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
         // Step 1c: Figure out all adjacent labels
         if (i / image.width == 0) {
             // Top Row (1 other pixel)
-            if (auto left = componentPoints[i - 1]; left != 0) {
```
nguy8tri, line 282: I prefer the compactness here. It also scopes the variables properly, as opposed to pulling them out of the if

src/distance/edge.cpp
```diff
@@ -338,17 +367,17 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
     // label into the lower label, and update the lowest and highest points,
     // and then get rid of the higher label's component data. We iterate from highest to lowest
     for (int i = L; i >= 0; i--) {
-        auto it = equivalencies.find(i);
+        std::unordered_map<int, int>::iterator it = equivalencies.find(i);
```
nguy8tri, line 362: so here we choose to use the standard type?

src/distance/edge.cpp
```diff
@@ -361,9 +390,15 @@ Components ConnectedComponentsAlgorithm(const Image &image, std::function<bool(u
 
     // Step 3: Return the components
     Components result;
-    for (const auto &[label, component] : components) result.push_back(component);
```
nguy8tri, line 369: Its nice to destructure here, so try using the original type if you can.

src/providers/converters.hpp
```diff
@@ -77,12 +78,12 @@ inline EulerAngles strtoea(const std::string &str) {
     size_t index = 0;
 
     while (index != 2 && end != std::string::npos) {
-        result[index++] = strtodecimal(str.substr(start, end - start));
+        result[index++] = strtodecimal(str.substr(start, end - start));  // GCOVR_EXCL_BR_LINE
```
nguy8tri, line 81: Curious as to how these lines became uncovered, that doesn't make much sense

src/providers/factory.hpp
```diff
@@ -16,9 +16,10 @@ namespace found {
  * 
  * @return A pointer to a CalibrationPipelineExecutor
  */
-inline std::unique_ptr<CalibrationPipelineExecutor> CreateCalibrationPipelineExecutor(CalibrationOptions &&options) {
-    return std::make_unique<CalibrationPipelineExecutor>(std::forward<CalibrationOptions>(options),
-                                    ProvideCalibrationAlgorithm(std::forward<CalibrationOptions>(options)));
+inline CalibrationPipelineExecutorPtr CreateCalibrationPipelineExecutor(CalibrationOptions options) {
+    static pool<CalibrationPipelineExecutor, 1> pool;  // GCOVR_EXCL_BR_LINE
```
nguy8tri: Again, doesn't make sense why this is uncovered, is it the coverage wrong. Also, does passing an empty pool for the dynamic mode make sense?

src/common/etl_config.hpp
```diff
@@ -0,0 +1,21 @@
...
+#define SRC_COMMON_ETL_CONFIG_HPP_
+
+// Default max sizes for ETL containers
+#ifndef FOUND_MAX_POINTS
+#define FOUND_MAX_POINTS 4096
+#endif
+#ifndef FOUND_MAX_IMAGE_PIXELS
+#define FOUND_MAX_IMAGE_PIXELS 1048576
```
nguy8tri: You should know that the image pixels, if you have the 2 dimensions, allow you to deduce the max points, components and edges. Its a substantial number, but you should be able to construct a proof for it (I may have left something for it in the giant software documentation). This will reduce the number of macros needed to be defined.

I will give you a hint (mostly because I don't remember). The maximum number of components and pixels is roughly 1/4th of the total pixels in the image This is because connectedness takes 4 of the points around a given pixel. There is a more exact formula, but you should be able to deduce it. The question to ask yourself is how many components you can stuff into an image without them being considered the same. The maximum number of edges thus follows from this, but could be smaller depending on how many points must be in a component to be an edge

CMakeLists.txt
```diff
@@ -195,63 +229,95 @@ endif()
...
+set_target_properties(found-test PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${BIN_DIR})
+if(TEST_LIBS)
+    add_dependencies(found-test ${TEST_LIBS})
+endif()
+
 add_custom_target(test
                   DEPENDS found-test
                   COMMENT "Compiles test code")
 
+add_custom_target(
```
nguy8tri, line 307: Why make this seperate?

Makefile
```diff
@@ -263,4 +280,7 @@ release: OPTIMIZATION = -O3
 release: CXXFLAGS += -DNDEBUG
 release: compile
 
--include $(SRC_OBJS:.o=.d) $(TEST_OBJS:.o=.d)
\ No newline at end of file
+-include $(SRC_OBJS:.o=.d) $(TEST_OBJS:.o=.d)
+
+$(ETL_CACHE_DIR):
```
nguy8tri: Should be in the appropriate location, not below the include

Makefile
```diff
@@ -263,4 +280,7 @@ release: OPTIMIZATION = -O3
 release: CXXFLAGS += -DNDEBUG
 release: compile
 
--include $(SRC_OBJS:.o=.d) $(TEST_OBJS:.o=.d)
```
nguy8tri, line 281: The empty next line should be preserved

nguy8tri (changes requested):
Hi Mahir, great job on the first iteration of this, I mainly left comments surrounding how you've structured the usage of ETL here for this round. One main piece of feedback is that for unique pointers that are not also arrays, you can either make a pool in the main thread, or just rely on the heap. You should be able to figure out which one is better for the program. There's also an excessive use of ETL in a few places, like the DataFile object, where a heap allocated object will function better in almost any case.

I think you should do an evaluation of where it should be used and where it could be spared. I'd say the main algorithms definitely make sense for ETL when using extended data structures. We should also think about how we can make it flexible for build as well, for instance its not entirely obvious if I just use build-etl that that a user can just change the dimensions

PR #72, opened 2026-04-29
Calculating covariance from CRA

rwilliaise:
This adds a new function stage (`SpheroidDistanceAndCovarianceDeterminationAlgorithm`) that wraps around the `SpheroidDistanceDeterminationAlgorithm` to additionally calculate the covariance of the estimated position. Importantly this is not a `DistanceDeterminationAlgorithm`, instead returning a `DistanceAndCovariance` struct which contains both a estimate and it's covariance. This is a break from the pipeline model but is the best compromise for right now.

### DIFF
--- a/src/common/style.hpp
+++ b/src/common/style.hpp
@@ -22,6 +22,14 @@ typedef std::vector<Vec2> Points;
 /// coordinate system.
 typedef Vec3 PositionVector;
 
+/// The output of a DistanceAndCovarianceDeterminationAlgorithm. Outputs a 3D
+/// Vector and its associated covariance matrix that represents the satellite's
+/// position.
+struct DistanceAndCovariance {
+    PositionVector distance;
+    Mat3 covariance;
+};
+
 /**
  * Represents an image
  * 
--- a/src/distance/distance.cpp
+++ b/src/distance/distance.cpp
@@ -51,6 +51,51 @@ PositionVector SpheroidDistanceDeterminationAlgorithm::Run(const Points &p) {
     return vecToEarth;
 }
 
+///// SpheroidDistanceAndCovarianceDeterminationAlgorithm /////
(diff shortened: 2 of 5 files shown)


src/common/style.hpp
```diff
@@ -22,6 +22,14 @@ typedef std::vector<Vec2> Points;
 /// coordinate system.
 typedef Vec3 PositionVector;
 
+/// The output of a DistanceAndCovarianceDeterminationAlgorithm. Outputs a 3D
+/// Vector and its associated covariance matrix that represents the satellite's
+/// position.
+struct DistanceAndCovariance {
```
leoca42, line 28: Rename this struct to PositionAndCovariance?

src/distance/distance.cpp
```diff
@@ -51,6 +51,59 @@ PositionVector SpheroidDistanceDeterminationAlgorithm::Run(const Points &p) {
...
+        // p[i] = u_i
+        // x_i
+        const Vec3 &point = algorithm_->getCamera().PixelToImageCoordinates(p[i]);
+
+        // see Exercise 8.6
+        Mat3 pointCovariance = Vec3(pointVariance, pointVariance, 0).asDiagonal();
+
+        decimal residualVariance = 4.0 * point.transpose() * transformedConicLocusMatrix *
+            pointCovariance * transformedConicLocusMatrix * point;
+        Mat3 jacobian = 2.0 * vecToEarth.transpose() *
```
leoca42: This is not a Mat3

nguy8tri:
If it is ready, this should be pushed into review

rwilliaise:
This hasn't been pushed for review seeing as there are no tests written yet, but we're not fully sure how this system would be tested or *if* this system should have unit testing.

